Issue #89, opened 2017-09-27
VSTS authenticated feeds are not supported.


jwittner:
Right now there's no authentication process for getting credentials through the client into the api requests.

Can we run the CredentialProvider ourselves?
Can we use user name and PAT to auth?

ForrestTrepte:
It looks like I can authenticate with my VSTS feed as follows:
1) Add credentials to the HttpWebRequest e.g. getRequest.Credentials = new NetworkCredential("forrest@contoso.com", "_VSTS Token Goes Here_")
2) Add the package source under NuGetForUnity Preferences with the V2 base URL e.g. https://myproject.pkgs.visualstudio.com/_packaging/myproject/nuget/v2/.

I will work on adding a user name and password field to the NuGetForUnity preferences UI for this. Will make a pull request when I get this working.

ForrestTrepte:
I may have spoken too soon. Although this successfully downloads the list of packages from VSTS, it fails when attempting to install one of those packages. It fails with "The remote server returned an error: (400) Authentication information is not given in the correct format. Check the value of Authorization header.." Frustratingly, the same code that fails in Unity succeeds when run in a C# console application. Perhaps a difference in between Mono and .NET HttpWebRequest?

ForrestTrepte:
This was fixed by #96.

PR #96, opened 2017-10-06
authentication for Visual Studio Team Services NuGet feed

ForrestTrepte:
This addresses #89 by adding a password field to the package sources.


Assets/NuGet/Editor/NugetConfigFile.cs
```diff
@@ -76,6 +77,14 @@ public void Save(string filepath)
                     addElement.Add(new XAttribute("value", "true"));
                     disabledPackageSources.Add(addElement);
                 }
+
+                if (source.HasPassword)
+                {
+                    addElement = new XElement("add");
```
ForrestTrepte: Josh noticed that this config file is a standard. And this code here isn't the right format. This should be saved in packageSourceCredentials instead. I'll make that change.

Assets/NuGet/Editor/NugetConfigFile.cs
```diff
@@ -76,6 +77,14 @@ public void Save(string filepath)
...
+                    addElement.Add(new XAttribute("key", source.Name));
```
jwittner: Agreed - we should move this over to the standard.

Assets/NuGet/Editor/NugetConfigFile.cs
ForrestTrepte: Fixed.

Assets/NuGet/Editor/NugetConfigFile.cs
ForrestTrepte: Fixed.

Ziugy:
Just pulled this down and tested in a module I was created and worked as intended. Thank you!

ForrestTrepte:
@GlitchEnzo: Do you think that you will have time to review this and the other outstanding pull requests sometime soon?

GlitchEnzo:
@ForrestTrepte Yep! Sorry for the delay. I saw the pull requests coming in, but I've been really busy and haven't had a chance to look them over in detail. I'll get to them in a couple of hours.

ForrestTrepte:
@GlitchEnzo Thanks for the invitation to collaborate with Push permissions!

It would be cool if you have time to review these first pull couple requests to see if any concerns jump out at you. Or just let me know know if you are too busy and want me to go ahead and merge them in.

Going forward, it will be nice for me to be able to merge pull requests and take some of the weight off your shoulders if you are too busy.

GlitchEnzo:
@ForrestTrepte You're welcome!  I'll look over the existing pull requests, but I wanted to avoid another situation where you guys are waiting for me for a week or more, thus the permissions.

ProjectSettings/ProjectSettings.asset
```diff
@@ -521,7 +521,7 @@ PlayerSettings:
     iOS: 1
   incrementalIl2cppBuild: {}
   additionalIl2CppArgs: 
-  scriptingRuntimeVersion: 0
+  scriptingRuntimeVersion: 1
```
GlitchEnzo: > [.NET 4.6] framework will become the default in a forthcoming version of Unity.

Is this documented somewhere? I know the default is still .NET 3.5 in Unity 2017.2.

Prior to this change, NuGetForUnity only required .NET 3.5 and could be installed and used in a Unity 5.2 or later project.  With this change, the entire Unity 5.x line would be excluded as users would need Unity 2017.1 or later. Are we really okay with excluding those users?

Sidenote: I wish there was an official Unity version usage chart that showed where the majority of users were.  That'd really help with making decisions such as these.

ForrestTrepte: [This post](https://forum.unity.com/threads/future-plans-for-the-mono-runtime-upgrade.464327/) suggests soon, but apparently it wasn't ready for 2017.2. You raise a good point. Since this isn't the default I'll revert the runtime version update. The authentication feature won't work by default, but anyone who needs it can switch their project to 4.6.

Ziugy: From my experience installing packages had no issues being on .NET 3.5. Though, it was required for getting packages restored.

ForrestTrepte: When I pass authentication information, I have an issue on .NET 3.5 that seems to be fixed in .NET 4.6. I have reverted my change to the .NET version and added an error message to inform folks who experience the authentication issue that they may need to update to 4.6.

jwittner:
@GlitchEnzo I got an email for an invite too - thanks so much! Unfortunately the link to accept doesn't appear to be working. =(

ProjectSettings/ProjectSettings.asset
ForrestTrepte: @GlitchEnzo Please take a look at my latest commit when you get a chance. Should be quick to review.

PR #98, opened 2017-10-09
skip NugetWindow.OnEnable when entering play mode and changing scripts (assembly reload) to resolve #94

### LINKED ISSUE #94: automatic restore operations slow down development
Each time you hit Play, a NuGet restore operation is run. This can take a while, particularly on a machine running over WiFi. There may be other cases in addition to hitting play when restore is run.

In order to speed up development, I'm thinking of adding an option to preferences that disables automatically running the restore operation. A Restore button or menu item would be added that would explicitly trigger a restore.

Users could disable automatically running restores, resulting in faster development at the cost of possibly becoming out of date if you forget to manually run a restore.

ForrestTrepte:
This relates to #94, skipping updates from NuGetWindow.OnEnable when hitting Play.

### DIFF
--- a/Assets/NuGet/Editor/NugetHelper.cs
+++ b/Assets/NuGet/Editor/NugetHelper.cs
@@ -1208,10 +1208,7 @@
             {
                 if (stopwatch.ElapsedMilliseconds >= 750)
                 {
-                    //LogVerbose("Downloading image timed out! Took more than 750ms.");
-
                     request.Dispose();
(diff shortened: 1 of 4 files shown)


ForrestTrepte:
I have updated this pull request to also skip the time-consuming parts of NuGetWindow.OnEnable following a recompile.

Ziugy:
This looks like a great update, +1!

PR #106, opened 2017-10-18
support environment variables in NuGet.config values

### LINKED ISSUE #97: config file should support environment variables
The [config docs](https://docs.microsoft.com/en-us/nuget/schema/nuget-config-file) say "you can use environment variables in NuGet.Config values to apply machine-specific settings at run time". I would like to support this feature in NuGetForUnity. My intended use is in combination with pull request #96. In our Unity projects, we are planning to check in NuGet.config files pointing to our team's feed. However, we don't want to check in our passwords so I am planning to save those in an environment variable.

For example:
```
  <packageSourceCredentials>
    <TeamFeed>
      <add key="clearTextPassword" value="%TeamFeedPassword%" />
    </TeamFeed>
  </packageSourceCredentials>
```

ForrestTrepte:
This resolves #97. There are now two versions of these variables:
- Saved: This is the literal version containing the environment variables that is edited and saved into the config file.
- Expanded: This is the version with the environement variables expanded that is used to perform the actual NuGet operations.

### DIFF
--- a/Assets/NuGet/Editor/NugetConfigFile.cs
+++ b/Assets/NuGet/Editor/NugetConfigFile.cs
@@ -1,5 +1,6 @@
 ﻿namespace NugetForUnity
 {
+    using System;
     using System.Collections.Generic;
     using System.IO;
     using System.Linq;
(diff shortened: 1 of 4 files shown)


jwittner:
Pulled down and verified that NuGet still functions and that adding a VSTS feed set as %VSTS_Token% with VSTS_Token being a valid environment variable set with my PAT works for our VSTS feed.

PR #108, opened 2017-10-19
Provide option for using mozroots to import Mozilla certs, give details otherwise.

### LINKED ISSUE #85: Certificate Validation is bypassed for the web requests used.
Right now certificate validation is bypassed, which is a serious security concern. 

Currently thinking we could add support for warning the user and letting them opt-in to the bypass action (with optional future white-listing).

jwittner:
Resolves #85 in my opinion.

### DIFF
--- a/Assets/NuGet/Editor/NugetPackageSource.cs
+++ b/Assets/NuGet/Editor/NugetPackageSource.cs
@@ -8,6 +8,7 @@
     using System.Net;
     using System.Xml;
     using System.Xml.Linq;
+    using UnityEditor;
     using Debug = UnityEngine.Debug;
 
     /// <summary>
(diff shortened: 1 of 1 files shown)


Assets/NuGet/Editor/NugetPackageSource.cs
```diff
@@ -321,27 +322,25 @@ private List<NugetPackage> GetPackagesFromUrl(string url, string password)
...
                 foreach (var package in packages)
                 {
                     package.PackageSource = this;
                 }
             }
-            catch (System.Exception e)
+            catch (Exception e)
             {
+                var we = e as WebException;
+                if (we != null && we.Status == WebExceptionStatus.TrustFailure && TryHandleTrustFailure())
```
ForrestTrepte: With the .NET 3.5 runtime, this occurs with we.Status of SendFailure and we.Message of "Error getting response stream (Write: The authentication or decryption has failed.): SendFailure".

A good way to handle that case may be to have an IsTrustFailure method with a case for detecting the .NET 3.5 error by string in addition to the .NET 4.6 error by Status code.

Assets/NuGet/Editor/NugetPackageSource.cs
```diff
@@ -321,27 +322,25 @@ private List<NugetPackage> GetPackagesFromUrl(string url, string password)
 
             try
             {
-                // Mono doesn't have a Certificate Authority, so we have to provide all validation manually.  Currently just accept anything.
-                // See here: http://stackoverflow.com/questions/4926676/mono-webrequest-fails-with-https
-
-                // remove all handlers
-                ServicePointManager.ServerCertificateValidationCallback = null;
```
jwittner: There are more cases of this - See NugetHelper.cs line 1057.

jwittner:
I'm abandoning this.

PR #110, opened 2017-10-23
Fix for stack trace log type warning

jwittner:


Ziugy:
Thank you for cleaning up the warnings!

PR #111, opened 2017-10-23
Install via the identifier to fix VSTS

### LINKED ISSUE #107: Dependencies in VSTS are not made available.
The feed returns no dependencies `<d:Dependencies m:null="true" />` when using VSTS Nuget Feeds.

Full URL we're seeing this result with:
```"https://myteam.pkgs.visualstudio.com/_packaging/myfeed/nuget/v2/Search()?$filter=IsAbsoluteLatestVersion&$orderby=DownloadCount desc&$skip=0&$top=15&searchTerm=''&targetFramework=''&includePrerelease=true"```

jwittner:
VSTS doesn't include dependencies on multipackage queries. By installing
by package id we query explicitly for this single package and thus get
the dependencies. This works for nuget.org as well.

Resolve #107

### DIFF
--- a/Assets/NuGet/Editor/NugetWindow.cs
+++ b/Assets/NuGet/Editor/NugetWindow.cs
@@ -814,7 +814,7 @@
                     {
                         if (GUILayout.Button("Install", installButtonWidth))
                         {
-                            NugetHelper.Install(package);
+                            NugetHelper.InstallIdentifier(package);
                             AssetDatabase.Refresh();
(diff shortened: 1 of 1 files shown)


jwittner:
@ForrestTrepte - found a pretty small change to fix this. =)

ForrestTrepte:
I take it that this works because GetSpecificPackage forces it to have specifically loaded the package individually via an API for which VSTS returns complete dependency information. Is that the right idea?

PR #112, opened 2017-10-25
Case insensitive compare

### LINKED ISSUE #104: NugetHelper.Clean is case sensitve.
In NugetHelper.Clean directory names are searched for after being translated to all lowercase letters (hinting that the directory can be case insensitive) but the final check for selectedDirectories.Contains(directory.FullName) does not translate to lower case letters.

For example, a DLL in Assets/lib/unity/... works but in Assets/lib/Unity/... does not work.

jwittner:
Fixes #104

### DIFF
--- a/Assets/NuGet/Editor/NugetHelper.cs
+++ b/Assets/NuGet/Editor/NugetHelper.cs
@@ -394,7 +394,11 @@
                 // delete all of the libaries except for the selected one
                 foreach (var directory in libDirectories)
                 {
-                    if (!selectedDirectories.Contains(directory.FullName))
+                    bool validDirectory = selectedDirectories
+                        .Where(d => string.Compare(d, directory.FullName, ignoreCase: true) == 0)
+                        .Any();
+
+                    if (!validDirectory)
                     {
                         DeleteDirectory(directory.FullName);
                     }


Ziugy:
Looks good! Definitely ran into this issue and this PR would save hours of wondering what's going on.

Assets/NuGet/Editor/NugetHelper.cs
```diff
@@ -394,7 +394,11 @@ private static void Clean(NugetPackageIdentifier package)
                 // delete all of the libaries except for the selected one
                 foreach (var directory in libDirectories)
                 {
-                    if (!selectedDirectories.Contains(directory.FullName))
+                    bool validDirectory = selectedDirectories
```
ForrestTrepte: I think a simpler fix would be to modify the "unity", ""net35-unity full v3.5", and "net35-unity subset v3.5" case above. Notice how all the other parts of the else if do selectedDirectories.Add(directory.FullName). Because they are using the actual name of the directory it will match selectedDirectories.Contains when it gets here. It's just the "unity" cases that are hard-coding a lowercase string. Would it be possible to change that section to use the original directory names? Since it looks like the intent is to include multiple directories (unlike the other cases that choose a best single lib folder), I think it would need to perform a search of libDirectories selecting all of those that match. This would keep the logic more similar across the cases and then this part could remain a simple Contains check. Sorry to nitpick on this. @jwittner, what do you think?

jwittner: The hard part is that if I find any unity directory I want accept all the unity directories. It'll make the loop strange, but perhaps I can pull out the unity search and only perform the non-unity search afterward. I'll look at this soon.

PR #123, opened 2017-12-21
Updated NugetHelper to handle netstandard

HassleTauf:
https://docs.microsoft.com/en-us/dotnet/standard/net-standard

### DIFF
--- a/Assets/NuGet/Editor/NugetHelper.cs
+++ b/Assets/NuGet/Editor/NugetHelper.cs
@@ -93,7 +93,7 @@
             {
                 return;
             }
-            
+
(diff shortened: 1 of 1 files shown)


GlitchEnzo:
Thanks for the pull request and the interest!

We already have a branch that has preliminary .NET Standard support:
https://github.com/GlitchEnzo/NuGetForUnity/commit/dc420bec5d8ebaf4d5b97ce2929bdea34d80542a

As you can see, it currently enables .NET Standard via a checkbox in the preferences, but it should probably be updated to handle the new ApiCompatibilityLevel in Unity 2018.1

PR #140, opened 2018-03-12
config setting that installs package files as read-only

### LINKED ISSUE #137: option to make files in the Packages read-only to prevent accidentally editing them
We have been having a problem where developers may inadvertently edit source code files installed from a package. We'd like to add an option to the NuGetForUnity preferences so that all package files are set to read-only. Unless this would cause a problem, I'm thinking this option should be turned on by default.

In Visual Studio it's easy to go to the definition of a function and make a change without realizing that you are inside the Packages folder. Later, you may commit your work to other files without realizing that your change to the Packages folder isn't included (we have Packages in the ignore list) and then you'll be confused when your change doesn't work for everyone else on your team.

ForrestTrepte:
This resolves #137.

### DIFF
--- a/Assets/NuGet/Editor/NugetConfigFile.cs
+++ b/Assets/NuGet/Editor/NugetConfigFile.cs
@@ -45,6 +45,11 @@
         public bool InstallFromCache { get; set; }
 
         /// <summary>
+        /// Gets or sets a value indicating whether installed package files are set to read-only.
+        /// </summary>
+        public bool ReadOnlyPackageFiles { get; set; }
+
+        /// <summary>
         /// The incomplete path that is saved.  The path is expanded and made public via the property above.
         /// </summary>
         private string savedRepositoryPath;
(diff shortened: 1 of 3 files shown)


ForrestTrepte:
We can ignore the failed appveyor check. I believe this is an automated build system that is under construction and isn't expected to work yet.

ForrestTrepte:
I was wondering about the default too. As @PJBowron points out, defaulting to false would match previous functionality. On the other hand, my team has been having problems due to accidentally modifying package files and it looks to me as if NuGetForUnity is designed to get packages as-is without modifying them. (The readme recommends that the whole Packages folder can be ignored in version control.) If true is generally the better option, then defaulting to false would doom future users who don't know to switch that setting.

I decided to change behavior, turning on read-only for everyone, because I thought that would be helpful in most cases. But I could be convinced otherwise--what do others think? @GlitchEnzo, do you have feelings about this?

GlitchEnzo:
I'm totally fine with making installed packages read-only by default.

ForrestTrepte:
Thanks, @GlitchEnzo and @PJBowron!

PR #144, opened 2018-03-16
use legacy www instead of UnityWebRequest for compability with Unity 5.2

### LINKED ISSUE #143: build fails for Unity 5.2 due to call of UnityWebRequest
The NugetWindow.CheckForUpdates feature that I introduced in #136 fails to build in Unity 5.2 because UnityWebRequest doesn't yet exist in that version of Unity.

ForrestTrepte:
Fixes #143.

### DIFF
--- a/Assets/NuGet/Editor/NugetWindow.cs
+++ b/Assets/NuGet/Editor/NugetWindow.cs
@@ -211,8 +211,15 @@
         protected static void CheckForUpdates()
         {
             const string url = "https://github.com/GlitchEnzo/NuGetForUnity/releases";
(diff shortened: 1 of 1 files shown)


ForrestTrepte:
Hi @StephenHodgson! I'll do that, but we release a .unitypackage that contains built .dlls instead of source code. So that version would be built on Unity 5.2 and would rely on WWW even when it is run in later versions of Unity. I suppose it's a good idea to have the compiler directive in there, though, so that if we move past Unity 5.2 in the future it will switch over. I'll try that...

ForrestTrepte:
@jwittner and @StephenHodgson: I have updated with conditional compilation as suggested by Stephen.

jwittner:
Pulling down the packager version of unity now to test this in multiple unity versions.

jwittner:
Was able to build the dlls and test in Unity!

PR #147, opened 2018-04-06
Support for NuGet 3.3+ local feed hierarchial folders structure

SINePrime:
Exactly what the title says.


OriginalSINe:
This uses an absolutely absurd amount of RAM. Closing this and fixing it.

OriginalSINe:
There is a RAM usage issue, but the problem exists outside of this change.

OriginalSINe:
As it turns out, the RAM issue was caused by #161.

.gitignore
```diff
@@ -72,4 +72,5 @@ Packages.meta
 
 # Ignore the bin folder
 CreateDLL/bin
-*.suo
\ No newline at end of file
+*.suo
+**/.vs/
```
jwittner: Please drop this change.

Assets/NuGet/.vs/ProjectSettings.json
```diff
@@ -0,0 +1,3 @@
+{
```
jwittner: This folder shouldn't be added.

Assets/NuGet/.vs/VSWorkspaceState.json
```diff
@@ -0,0 +1,7 @@
+{
```
jwittner: This folder shouldn't be added.

Assets/NuGet/Editor/NugetPackageSource.cs
```diff
@@ -556,5 +630,14 @@ private List<NugetPackage> GetUpdatesFallback(IEnumerable<NugetPackage> installe
 
             return updates;
         }
+
+        /// <summary>
+        /// Get if this <see cref="IsLocalPath"/> and the version is 3.3 or higher;
+        /// </summary>
+        private bool getIsLocalPathAndVersion33OrHigher()
```
jwittner: Should be PascalCase.

Assets/NuGet/Editor/NugetHelper.cs
```diff
@@ -304,7 +304,7 @@ private static void Clean(NugetPackageIdentifier package)
             if (Directory.Exists(packageInstallDirectory + "/lib"))
             {
                 int intDotNetVersion = (int)DotNetVersion; // c
-                //bool using46 = DotNetVersion == ApiCompatibilityLevel.NET_4_6; // NET_4_6 option was added in Unity 5.6
+                                                           //bool using46 = DotNetVersion == ApiCompatibilityLevel.NET_4_6; // NET_4_6 option was added in Unity 5.6
```
jwittner: nit: what's happened here?

Assets/NuGet/Editor/NugetPackageSource.cs
```diff
@@ -77,6 +77,11 @@ public bool HasPassword
         /// </summary>
         public bool IsLocalPath { get; private set; }
 
+        /// <summary>
+        /// Get or sets a value indicating if this is a local path that is 3.3+;
+        /// </summary>
+        public bool IsLocalPathAndVersion33 { get; set; }
```
jwittner: Should be a `private set`, and I'd recommend making this just track the IsVersion33 test. As implemented this can diverge from 'IsLocalPath'. You could write a property to return the result of these two state properties.

JoC0de:
Implemented in #489. I solved it the pragmatic way.

PR #156, opened 2018-05-04
.NET Standard 2.0 (Support Unity 2018)

OriginalSINe:
There's several branches for this, but none have a pull request.


jwittner:
Tested in Unity 2017.3.1f1 and Unity 2018.1.3f1 and it worked well!

PR #160, opened 2018-05-13
Include package /content files

### LINKED ISSUE #159: Include package files under /content
## Description
Currently, packaged files under /content are deleted in NugetHelper.Clean. This does not seem like the desired behavior.

As a sidenote, files under /contentFiles are kept. (Seems to contain the same files?)

- **NuGet Package:** N/A.
- **NuGetForUnity Version:** 0.15
- **Unity Version:** N/A.
- **Operating System:** N/A.

OriginalSINe:
Fixes #159


jwittner:
I've wondered about this too - seems much better to keep it.

PR #164, opened 2018-06-12
Improve VSTS nuget feed support

CoPrez:
Visual Studio Team Services hosted projects have built in support for hosting nuget feeds, but authenticating with them is a bit cumbersome. NuGetForUnity currently supports connecting to the feed by generating a personal access token (PAT) and passing that in using basic HTTP authentication. Howerver that requires each user to go out of their way to generate a PAT and get it piped into NuGetForUnity.

This change adds support for using CredentialProvider.VSS.exe to obtain the authentication token rather than requiring the user to generate it (see [here](https://docs.microsoft.com/en-us/vsts/package/nuget/nuget-exe?view=vsts#download-the-credential-provider-directly) for more info on the credential provider). The first time the user connects to a VSTS feed they will get a prompt to enter their VSTS credentials (through the CredentialProvider.VSS.exe). Subsequent connections after that will be silent as the credential provider handles caching the token as well.

### DIFF
--- a/Assets/NuGet/Editor/NugetHelper.cs
+++ b/Assets/NuGet/Editor/NugetHelper.cs
@@ -1169,7 +1169,7 @@
                 getRequest.ReadWriteTimeout = timeOut.Value;
             }
 
-            if (packageHost.EndsWith("pkgs.visualstudio.com") && password == null)
+            if (packageHost.EndsWith("pkgs.visualstudio.com") && string.IsNullOrEmpty(password))
             {
                 // The host is a VisualStudio feed (which requires authentication) but a password was not provided. Use the VSS credential provider to aquire a token and append
                 // it to the request.
--- a/Assets/NuGet/Editor/NugetHelper.cs
+++ b/Assets/NuGet/Editor/NugetHelper.cs
@@ -1315,6 +1315,12 @@
             public string Password;
         }
 
+        /// <summary>
+        /// Helper function to aquire a token to access VSTS hosted nuget feeds by using the CredentialProvider.VSS.exe
+        /// tool. Downloading it from the VSTS instance if needed.
+        /// </summary>
+        /// <param name="packageHost">The hostname where the VSTS instance is hosted (such as microsoft.pkgs.visualsudio.com</param>
+        /// <returns>The password in the form of a token, or null if the password could not be aquired</returns>
         private static string GetPasswordFromVSTSCredentialProvider(string packageHost)
         {
             string credentialProviderBundleFilename = "CredentialProviderBundle.zip";
--- a/Assets/NuGet/Editor/NugetHelper.cs
+++ b/Assets/NuGet/Editor/NugetHelper.cs
@@ -1304,20 +1304,32 @@
         }
 
         /// <summary>
-        /// Data class returned from the CredentialProvider.VSS.exe tool. The tool returns this datatype as a
-        /// JSON encoded string which can be decoded using Unity's JsonUtility.
+        /// Data class returned from nuget credential providers in a JSON format. As described here:
+        /// https://docs.microsoft.com/en-us/nuget/reference/extensibility/nuget-exe-credential-providers#creating-a-nugetexe-credential-provider
         /// </summary>
         [System.Serializable]
-        private struct VSSCredentialResponse
+        private struct CredentialProviderResponse
         {
-            public string __VssPasswordWarning;
             public string Username;
             public string Password;
         }
 
         /// <summary>
+        /// Possible response codes returned by a Nuget credential provider as described here:
+        /// https://docs.microsoft.com/en-us/nuget/reference/extensibility/nuget-exe-credential-providers#creating-a-nugetexe-credential-provider
+        /// </summary>
+        private enum CredentialProviderExitCode
+        {
+            Success = 0,
+            ProviderNotApplicable = 1,
+            Failure = 2
+        }
+
+        /// <summary>
         /// Helper function to aquire a token to access VSTS hosted nuget feeds by using the CredentialProvider.VSS.exe
         /// tool. Downloading it from the VSTS instance if needed.
+        /// See here for more info on nuget Credential Providers:
+        /// https://docs.microsoft.com/en-us/nuget/reference/extensibility/nuget-exe-credential-providers
         /// </summary>
         /// <param name="packageHost">The hostname where the VSTS instance is hosted (such as microsoft.pkgs.visualsudio.com</param>
         /// <returns>The password in the form of a token, or null if the password could not be aquired</returns>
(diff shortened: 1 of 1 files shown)


Assets/NuGet/Editor/NugetHelper.cs
```diff
@@ -1291,5 +1302,80 @@ public static Texture2D DownloadImage(string url)
...
+        /// <param name="packageHost">The hostname where the VSTS instance is hosted (such as microsoft.pkgs.visualsudio.com</param>
+        /// <returns>The password in the form of a token, or null if the password could not be aquired</returns>
+        private static string GetPasswordFromVSTSCredentialProvider(string packageHost)
+        {
+            string credentialProviderBundleFilename = "CredentialProviderBundle.zip";
+            string credentialProviderFilename = "credentialprovider.vss.exe";
+            string credentialProviderPath = Path.Combine(NugetConfigFile.RepositoryPath, credentialProviderFilename);
+
+            // Download CredentialProvider if it's not already installed
+            if (!File.Exists(credentialProviderPath))
```
Ziugy: Maybe the first check that should be done is to see if the credential provider path is found in an environment variable?

Assets/NuGet/Editor/NugetHelper.cs
```diff
@@ -1291,5 +1302,80 @@ public static Texture2D DownloadImage(string url)
...
+            {
+                HttpWebRequest credentialProviderRequest = (HttpWebRequest)WebRequest.Create("https://" + packageHost + "/_apis/public/nuget/client/CredentialProviderBundle.zip");
+                Stream credentialProviderDownloadStream = credentialProviderRequest.GetResponse().GetResponseStream();
+
+                using (Stream file = File.Create(credentialProviderBundleFilename))
+                {
+                    CopyStream(credentialProviderDownloadStream, file);
+                }
+
+                // Unzip the bundle and extract thje credential provider executable
```
Ziugy: Nit: small typo - thje to the

Assets/NuGet/Editor/NugetHelper.cs
Ziugy: I would suggest we first check to see if the credential provider is within an environment variable and then download as the fallback.

Assets/NuGet/Editor/NugetHelper.cs
Ziugy: Nit: typo - thje to the

Assets/NuGet/Editor/NugetHelper.cs
```diff
@@ -1291,5 +1302,80 @@ public static Texture2D DownloadImage(string url)
...
+            process.StartInfo.CreateNoWindow = true;
+            process.StartInfo.RedirectStandardOutput = true;
+            process.StartInfo.FileName = credentialProviderPath;
+            process.StartInfo.Arguments = "-uri " + "https://" + packageHost;
+
+            // http://stackoverflow.com/questions/16803748/how-to-decode-cmd-output-correctly
+            // Default = 65533, ASCII = ?, Unicode = nothing works at all, UTF-8 = 65533, UTF-7 = 242 = WORKS!, UTF-32 = nothing works at all
+            process.StartInfo.StandardOutputEncoding = Encoding.GetEncoding(850);
+            process.Start();
+            process.WaitForExit();
```
Ziugy: Not sure how long this takes. Maybe we should show "thinking" status through a progress dialog?

CoPrez: The credential provider should only get hit at points where the nuget feed itself is being hit. I think most of those already have a progress bar dialog shown.

Assets/NuGet/Editor/NugetHelper.cs
CoPrez: Updated to match the search pattern that nuget itself uses as defined [here](https://docs.microsoft.com/en-us/nuget/reference/extensibility/nuget-exe-credential-providers#creating-a-nugetexe-credential-provider)

Assets/NuGet/Editor/NugetHelper.cs
CoPrez: fixed

Assets/NuGet/Editor/NugetHelper.cs
```diff
@@ -1291,5 +1302,113 @@ public static Texture2D DownloadImage(string url)
...
+        private static string GetPasswordFromVSTSCredentialProvider(string packageHost)
+        {
+            string credentialProviderBundleFilename = "CredentialProviderBundle.zip";
+            string credentialProviderFilename = "credentialprovider.vss.exe";
+
+            // Build the list of possible locations to find the credential provider. In order it should be local app data, paths set on the
+            // environment varaible, and lastly look at the root of the pacakges save location.
+            List<string> possibleCredentialProviderPaths = new List<string>();
+            possibleCredentialProviderPaths.Add(Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData), "Nuget", "CredentialProviders"));
+            possibleCredentialProviderPaths.AddRange(Environment.GetEnvironmentVariable("NUGET_CREDENTIALPROVIDERS_PATH")?.Split(';') ?? new string[]{});
```
jwittner: I get this error from Unity 2017.3.1f1 - the default for the project is 2017.1.1f1, but the DLL must be backwards compatible to `5.2.0f3` which is used by the packager. Curious what version you used.

```
Assets/NuGet/Editor/NugetHelper.cs(1342,122): error CS1644: Feature `null propagating operator' cannot be used because it is not part of the C# 4.0 language specification
```

CoPrez: I was testing in 2017.4.1f1, the real issue though was that I was using scripting back-end .NET 4.6. Updated the code and tested with .NET 3.5 as the backend and things seem to be working there. I don't have Unity 5.2.0f3 on had to test with.

Assets/NuGet/Editor/NugetHelper.cs
```diff
@@ -1160,19 +1160,30 @@ private static void WarnIfDotNetAuthenticationIssue(Exception e)
...
+            string packageHost = new Uri(url).Host;
+
             HttpWebRequest getRequest = (HttpWebRequest)WebRequest.Create(url);
             if (timeOut.HasValue)
             {
                 getRequest.Timeout = timeOut.Value;
                 getRequest.ReadWriteTimeout = timeOut.Value;
             }
+
+            if (packageHost.EndsWith("pkgs.visualstudio.com") && password == null)
```
Ziugy: Small change.

We should change `password == null` to `string.IsNullOrEmpty(password)`. They might have the password box ticked and nothing provided.

CoPrez: Good point! I'll get that updated.

Assets/NuGet/Editor/NugetHelper.cs
```diff
@@ -1291,5 +1302,119 @@ public static Texture2D DownloadImage(string url)
...
+            // http://stackoverflow.com/questions/16803748/how-to-decode-cmd-output-correctly
+            // Default = 65533, ASCII = ?, Unicode = nothing works at all, UTF-8 = 65533, UTF-7 = 242 = WORKS!, UTF-32 = nothing works at all
+            process.StartInfo.StandardOutputEncoding = Encoding.GetEncoding(850);
+            process.Start();
+            process.WaitForExit();
+
+            string output = process.StandardOutput.ReadToEnd();
+            if ((CredentialProviderExitCode)process.ExitCode == CredentialProviderExitCode.Success && !string.IsNullOrEmpty(output))
+            {
+                CredentialProviderResponse response = JsonUtility.FromJson<CredentialProviderResponse>(output);
```
jwittner: Hey! Working well from within the main project, but for packaging the CreateDLL project has to build against the version used by the packager - 5.2.0f3. Unfortunately I'm seeing this from that version:
```
 error CS0103: The name 'JsonUtility' does not exist in the current context
```
We've got a build script internally that builds unity packages from this repo. It wasn't included originally because UnitySetup wasn't OSS at the time. I think now I'll make a PR to add that script to this repo so it's easier to test.

jwittner: @GlitchEnzo - how do you feel about moving up our minimum supported version? Any idea how many people are still in the 5.x?

GlitchEnzo: It depends upon how far up we'd be moving the minimum.  The last metrics I saw (which are quite old) show a very low 5.2 usage.  I do know of a "major" customer that is still using 5.6, though. Some of the big companies are much slower in their upgrading.

The current version is still available to folks, so I won't push back too hard if you need to upgrade to an even higher minimum to help keep the development easier. If they desperately need old Unity support, they can always fork it themselves.

awolowiecki720:
Hey, I am taking these changes and I am trying to upgrade to the minimum supported Unity version that still works with jsonutility, which looks like it's ~5.3. Then I will submit a new PR.

jwittner:
With @awolowiecki720 update to the version in now, this is building. Approved, works great @CoPrez . Thanks again for the contributions!

Issue #171, opened 2018-07-02
Duplicate dependencies causing compilation errors


SINePrime:
## Description
When a package has a dependency on an assembly already present in Unity, compilation errors arise from duplicate assemblies.

Some such errors:
- Unity error CS1704
- The type "..." exists in both '...' and '...'

**NuGet Package:** QuickGraph.All
**NuGetForUnity Version:** 0.0.16
**Unity Version:** 2018.1.1f1
**Operating System:** Windows 10

jwittner:
I believe this has been fixed by #303

PR #172, opened 2018-07-02
Added support for "ignored" dependencies in config file

SINePrime:
A quick fix for Issue #171.

Ideally, ignored packages should be linked to the engine or editor. After installing, the "platform settings" should be programmatically changed to exclude that platform. This is because assemblies compiled for the engine and editor platforms have different default references.

ie, if a "plugin" included by default in engine platform, its "platform settings" should exclude the engine platform.

### DIFF
--- a/Assets/NuGet/Editor/NugetConfigFile.cs
+++ b/Assets/NuGet/Editor/NugetConfigFile.cs
@@ -50,6 +50,12 @@
         public bool ReadOnlyPackageFiles { get; set; }
 
         /// <summary>
+        /// Package IDs of dependancies to ignore.
+        /// ie, ones to not install;
+        /// </summary>
+        public HashSet<string> IgnoredDependencies { get; set; }
+
+        /// <summary>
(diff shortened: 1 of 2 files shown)


Assets/NuGet/Editor/NugetConfigFile.cs
```diff
@@ -306,11 +334,166 @@ public static NugetConfigFile CreateDefaultFile(string filePath)
...
 
             return Load(filePath);
         }
+
+        /// <summary>
+        /// Get all the dependencies that are referenced in Unity assemblies by default;
+        /// </summary>
+        private static IEnumerable<string> dependenciesUnityGet()
+        {
+            return new HashSet<string>(){
```
ForrestTrepte: How was this list generated? I'm concerned that it could be difficult to maintain this data. Would it be possible for this function to dynamically generate this by reflecting on Unity's assemblies? If not, can we document a set of steps for updating the list?

jwittner:
I believe recent changes have shifted the landscape on this problem.

Issue #165, opened 2018-06-14
Idea: Integration with Unity Package Manager (UPM)


capyvara:
## Description

On Unity 2017+ there's the UPM that already allows partial support for local client side controlled packages.

#### Some references:

[UPM: How to make a custom package](https://gist.github.com/LotteMakesStuff/6e02e0ea303030517a071a1c81eb016e)

[Local client side packages?](https://forum.unity.com/threads/local-client-side-packages.532521/)

### Idea outline
- Put the config files at `Packages/...` instead of `Assets/...`
- Install the packages directly into `Packages/...` instead of `Assets/Packages/...`
- Dynamically create a `Packages/Name/package.json` with the info from NuGet package

Or alternatively (not sure it this will work):
- Install the packages at unity shared cache folder (on mac for example is `~/Library/Unity/cache/packages/packages.unity.com/name@version/`) 
- Modify `Packages/manifest.json` dependencies and pray for Unity to find it

Or (not sure if will work too):
- Let NuGet install to the [default](https://docs.microsoft.com/en-us/nuget/consume-packages/managing-the-global-packages-and-cache-folders) cache path 
- Point to it using the file: scheme inside the `manifest.json`

jwittner:
This is not recommended by Unity here in the upgrade guide for the Package Manager drop: https://docs.unity3d.com/Manual/UpgradeGuide20181.html

Issue #133, opened 2018-02-27
Upcoming conflict with Unity Package Manager


jwittner:
Interesting note in the 2018.1.0b9 Unity Beta Release:

> Package Manager: Unity Package Manager project sub-folder was moved from UnityPackageManager to Packages. This is to align the physical path with the logical path of packages. Notes:
> * Projects configured with a NuGet should change their configuration to point to another folder (default repository path for NuGet configuration is \packages). The risk of NuGet and Unity Package Manager sharing the Packages folder is minimal, but we suggest to separate them nonetheless to avoid any issues.
> * When opening a project created with a previous version, a migration flow will copy the content of the UnityPackageManager folder to the Packages folder. It is left to the project maintainer to erase the UnityPackageManager folder and update source control configuration.

I suspect we're ok because we're putting the packages folder inside of Assets. I'm not sure what the UnityPackageManager has as an end goal.

MorphiksBuilder:
Well if they handle NuGet packages then it will be a relief and you could redirect your energy to other projects ?

Nothing says they're gonna handle nuget for now though

GlitchEnzo:
It would be great if the Unity Package Manager supported NuGet packages, but they are using completely different tech and I'm doubtful they would ever add support for NuGet.

https://docs.unity3d.com/Packages/com.unity.package-manager-ui@1.8/manual/index.html

jwittner:
Uh oh: 
![image](https://user-images.githubusercontent.com/12958442/38217463-57a8fdfe-3683-11e8-84b6-d9cda9ef4c9d.png)

GlitchEnzo:
While the DLLs and such are installed into (project)/Assets/Packages, the extra stuff, like documentation and PowerShell scripts are installed into (project)/Packages. So we will likely have to rename that outer folder.

aalmada:
2018.1.0f2 is officially available and it shows a warning when it finds the Packages folder inside Assets.

twhittaker:
Unity Hub version 0.20.1 now automatically relocates the package folder:
 "Unity needs to relocate your Packages folders to the UnityPackageManager folder under your Project.  You can safely delete the old folder after proceeding if it's no longer needed"

PR #190, opened 2018-10-09
Deleting unnecessary folders after update packages.config from git.

vandelpal:
Description of the problem that is solved:
When one developer uninstall any package and commit it. The folder remove on this developer pc, but other developer pull this changes, Assets/Packages folder in ignore, open Unity and automatically start restore packages. Folder still exist because there is no code for check this.

### DIFF
--- a/Assets/NuGet/Editor/NugetHelper.cs
+++ b/Assets/NuGet/Editor/NugetHelper.cs
@@ -1267,6 +1267,8 @@
 
                     currentProgress += progressStep;
                 }
+
+                CheckForUnnecessaryPackages();
             }
             catch (Exception e)
             {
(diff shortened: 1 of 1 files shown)


jwittner:
The issue I have with this operation is that it assumes NuGet owns everything in the repository. We often find it useful to embed an installed package into our source tree and remove it from the packages config - this change would prevent that work flow.

I'd be open to adding a 'Clean' option via another button or something similar to invoke the wipe then Restore. For those of us using the above workflow we'll have to use 'git' externally to restore that missing content, but at least the normal flow wouldn't break.

Open to other ideas as well, of course. =)

Visual Studio doesn't have this problem, because it uses a reference in the project to link to the package contents. We don't have that option in Unity right now. =(

vandelpal:
Nuget owns everything only in the Asset/Packages ignored folder, why not?  You can embed some package in other folder for source tree.

vandelpal:
In Unity 2018 the Packages folder is now used for its own purposes.  
From upgrade Guide:
"If your project uses Nuget (or any external package manager) in a way that makes it use the Packages directory, its configuration should be changed to use a different directory. This is recommended to eliminate the small chance that a package be picked up by the Unity Package Manager, which could result in hard-to-debug issues like compilation errors or import errors."

Therefore, I propose to change the folder and give it completely under the control of the Nuget.

jwittner:
The Unity Package Manager 'Packages' folder resides alongside the Assets folder, where as the NuGetForUnity 'Packages' folder is embedded inside the Assets folder. See #133 and #165 for references to this. =)

Seems reasonable to move the contents out of the packages folder. I'll test your PR soon and give it a functionality review.

PR #191, opened 2018-10-11
Cache icons in Application.temporaryCachePath folder

vandelpal:


mister-liss:
@jwittner I've pulled this down and verified it. I was able to find the cached icon for bootstrap, deface it and see the result. I also deleted the cache and saw the icons restored. Can we get this merged?

Assets/NuGet/Editor/NugetHelper.cs
```diff
@@ -1322,26 +1329,54 @@ public static Texture2D DownloadImage(string url)
...
+        }
+
+        private static string GetFilePath(string url) {
+            return Path.Combine(Application.temporaryCachePath, GetHash(url));
+        }
+
+        private static string GetHash(string s) {
+            if (string.IsNullOrEmpty(s))
+                return null;
+            MD5CryptoServiceProvider md5 = new MD5CryptoServiceProvider();
```
jwittner: Would love if the cache entries had an expiry associated with them so that server updates would eventually be captured.

Assets/NuGet/Editor/NugetHelper.cs
```diff
@@ -1311,6 +1312,12 @@ public static Texture2D DownloadImage(string url)
             Stopwatch stopwatch = new Stopwatch();
             stopwatch.Start();
 
+            bool fromCache = false;
+            if (ExistsInDiskCache(url)) {
+                url = "file:///" + GetFilePath(url);
```
jwittner: Can this be achieved via System.Uri?

Assets/NuGet/Editor/NugetHelper.cs
jwittner: Or you might try a HEAD query against the Url to compare the Entity properties.

mister-liss:
@jwittner Builds successfully. Good to go.

PR #197, opened 2018-10-26
Add MacOS support

suntabu:


Assets/NuGet/Editor/NugetHelper.cs
```diff
@@ -208,6 +211,21 @@ private static void RunNugetProcess(string arguments, bool logOuput = true)
 
             LogVerbose("Running: {0} \nArgs: {1}", files[0], arguments);
 
+#if UNITY_EDITOR_OSX
+            var monoFile = "/Library/Frameworks/Mono.framework/Versions/Current/Commands/mono";
+            //TODO:using unity mono to pack the Nuget package.
+            var commandLine = " " + files[0] + " " + arguments;
```
jwittner: Seems like more of this code could be shared between the two platforms. Looks like only the `FileName` and `Arguments` are different?

Assets/NuGet/Editor/NugetHelper.cs
```diff
@@ -36,8 +36,11 @@ public static class NugetHelper
         /// <summary>
         /// The path where to put created (packed) and downloaded (not installed yet) .nupkg files.
         /// </summary>
+#if UNITY_EDITOR_OSX
+        public static readonly string PackOutputDirectory = Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData) + "/NuGet/Cache";
```
jwittner: Would `Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData),"NuGet","Cache")` work on all platforms?

Assets/NuGet/Editor/NugetHelper.cs
```diff
@@ -208,6 +211,21 @@ private static void RunNugetProcess(string arguments, bool logOuput = true)
 
             LogVerbose("Running: {0} \nArgs: {1}", files[0], arguments);
 
+#if UNITY_EDITOR_OSX
+            var monoFile = "/Library/Frameworks/Mono.framework/Versions/Current/Commands/mono";
+            //TODO:using unity mono to pack the Nuget package.
```
jwittner: Can you drop this TODO and leave a comment that we're using mono to run 'nuget.exe' - this method is used for all nuget interfacing - not just to pack things.

Assets/NuGet/Editor/NugetHelper.cs
suntabu: Done

Assets/NuGet/Editor/NugetHelper.cs
suntabu: Yes, but 2 Path.Combine invokes would be used

Assets/NuGet/Editor/NugetHelper.cs
suntabu: Done

Assets/NuGet/Editor/NugetHelper.cs
jwittner: That seems ok right?

Assets/NuGet/Editor/NugetHelper.cs
jwittner: Yay!

Assets/NuGet/Editor/NugetHelper.cs
suntabu: Yes, only 2 arguments needed

jwittner: Thanks for the change! =)

Issue #207, opened 2018-11-15
Packages that fail to install dependencies should fail to install


derekfreed:
If I install a package with a dependency and the dependency fails to install, the package should not be put into my project. It should just get an error that it failed because it couldn't install dependencies.

ForrestTrepte:
@derekfreed, it looks like this was fixed by #208? Can we close this issue?

derekfreed:
Yup, meant to do that

PR #208, opened 2018-11-15
Makes package fail install if dependencies fail to install

derekfreed:
Now packages will fail to install if their dependencies fail to install


derekfreed:
Resolves issue [#207](https://github.com/GlitchEnzo/NuGetForUnity/issues/207)

jwittner:
Great change!

Issue #209, opened 2018-11-20
Package manager installs old version of package


Klon160:
## Description
I'm trying to install CsvHelper but after successful installation the whole package manager stops working (Error: Object reference not set to an instance of an object). Other packages (e.g. JSON.net) are not causing this error.
I tried a different version of NuGetForUnity (0.0.18) and got the same error. In both cases it installed the oldest version of CsvHelper "0.12.0" (released 2011) instead of "9.2.0".

- **NuGet Package:** [What is the name and version of the NuGet package you are installing?]
CsvHelper

- **NuGetForUnity Version:** [What version of NuGetForUnity are you running?]
1.0.0

- **Unity Version:** [What version of Unity are you using?]
2018.2.16f1

- **Operating System:** [What OS are you on?]
Windows 10

Screenshots:
![2018-11-20_16-03-49](https://user-images.githubusercontent.com/4398383/48783219-260cc280-ece0-11e8-9ad7-cdeca50a9cd2.png)
![unity_2018-11-20_16-05-35](https://user-images.githubusercontent.com/4398383/48783235-32911b00-ece0-11e8-9357-4d39c7035658.png)

Edit:
I even tried changing the package version manually in the config file, but after the changes I had an infinite loop where the package manager tries to download version 9.2.0, doesn't find it, downloads version 0.12.0, finds out that it should use a newer version, deletes 0.12.0 and tries to download 9.2.0 again.

bleppie:
I'm getting the same issue with Serilog, I hit install and get version 0.1.6 ratehr than 2.7.0. HIttting the Update button afterwards does nothing. Error:

Could not find Serilog [2.5.0, ) or greater.
UnityEngine.Debug:LogErrorFormat(String, Object[])
NugetForUnity.NugetHelper:InstallIdentifier(NugetPackageIdentifier, Boolean)
NugetForUnity.NugetHelper:Install(NugetPackage, Boolean)
NugetForUnity.NugetHelper:InstallIdentifier(NugetPackageIdentifier, Boolean)
NugetForUnity.NugetWindow:DrawPackage(NugetPackage, GUIStyle, GUIStyle)
NugetForUnity.NugetWindow:DrawPackages(List`1)
NugetForUnity.NugetWindow:DrawOnline()
NugetForUnity.NugetWindow:OnGUI()
UnityEngine.GUIUtility:ProcessEvent(Int32, IntPtr)

derekfreed:
[...]

PR #213, opened 2018-12-05
Fix bugs in GetOnlinePackage

derekfreed:
Was seeing cases where my package would be found in one source then overwritten by a lower version found on a different source.

May resolve issue [#209](https://github.com/GlitchEnzo/NuGetForUnity/issues/209)


jwittner:
Checking locally.

jwittner:
Worked great even in some of the odd multi-listing cases. =)

PR #221, opened 2018-12-27
NuGet.Unity DLL generation, UnityPackage support & Quality of Life Changes

dklompmaker:
Greetings,

We have been developing a pipeline that uses NuGet packages as well, and wanted to add some feature support to NuGetForUnity that we'd like to see.

In this pull-request you will find the following changes:

**1. Loose Code into Dlls**
To ease the implementation into projects, we've created a *.sln and *.csproj for both the core codebase and the unit-tests. The build output directory is targeting the sample Unity project; so you can simply click build to push the dll into the project. This change will allow the creation of NuGet.Unity as its own package for distribution as well; allowing developers to update NuGetForUnity via Nuget.

**2. UnityPackage Support**
Our development pipeline sees the use of *.unitypackage distributed within NuGet packages. This allows meta files to be maintained between projects, along-with the ability for developers to move assets where they like and maintain update support. The feature addition, gives users control to automatically import these *.unitypackages (if found within the extracted NuGet package).

Upon uninstallation of a NuGet package, we extract the contents of any *.unitypackages to find meta files. We use these meta-files to uninstall the assets wherever they are located in the project.

This is driven by the new configuration key `autoImportUnityPackages` in the `NuGet.config`. You can also edit this value in the preferences pane.

**3. Quality of Life (GUI Updates)**
We're going to continue to work in this area, but some of the changes already are:
1. Search bar & header that mimics Unity's styles.
2. Moving the menu-items under Window/ to better organize. 

**4. Version Meaning**
We use a three dot notation (similar to semantic) that represents: `{conceptual}.{major}.{minor}` in this scheme, each segment has the following meaning:

*conceptual*: Represents a potential re-write, with large refactor possible.
*major*: Any non-backwards compatible change, medium refactor possible.
*minor*: Backwards compatible changes, no refactor; includes builds.

We appreciate the great work done on this, and hope that we can continue to add value.

Thanks!


jwittner:
Some of this looks really great and quite useful, but this PR is too large to consider en masse and recent changes have shifted the landscape on some of the problem. Would love to consider the value of your ideas above in separate issues if you're still keen to contribute them.

PR #226, opened 2019-01-18
Async load images for packages and preview icon when create new package

vandelpal:
NugetWindow not wait while loading all images, not need magic timeout 750ms for www request.  NugetHelper is too big, this change reduces its size.
When create new Package, it will be convenient to immediately see a preview of the icon.


JoC0de:
Implemented in #483 see https://github.com/GlitchEnzo/NuGetForUnity/blob/620a6ce1e339db0ccc4048167bb87a3a8e0b0a8e/src/NuGetForUnity/Editor/NugetHelper.cs#L1651

PR #239, opened 2019-04-04
Delete runtimes folder from packages until better support can be added

jwittner:
Packages that have native dlls put them into runtime folders that are selected at runtime or when you elect a target platform. For Unity without removing this folder we end up with duplicate dlls that cause errors. I'm removing the directory for now until better support can come in for supporting and configuring the platform dlls in Unity native manners.

### DIFF
--- a/Assets/NuGet/Editor/NugetHelper.cs
+++ b/Assets/NuGet/Editor/NugetHelper.cs
@@ -311,6 +311,10 @@
             // For now, delete src.  We may use it later...
             DeleteDirectory(packageInstallDirectory + "/src");
 
+            // Since we don't automatically fix up the runtime dll platforms, remove them until we improve support
+            // for this newer feature of nuget packages.
+            DeleteDirectory(Path.Combine(packageInstallDirectory, "runtimes"));
+
             // Delete documentation folders since they sometimes have HTML docs with JavaScript, which Unity tried to parse as "UnityScript"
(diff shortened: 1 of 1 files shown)


Assets/NuGet/Editor/NugetHelper.cs
```diff
@@ -311,6 +311,10 @@ private static void Clean(NugetPackageIdentifier package)
             // For now, delete src.  We may use it later...
             DeleteDirectory(packageInstallDirectory + "/src");
 
+            // Since we don't automatically fix up the runtime dll platforms, remove them until we improve support
+            // for this newer feature of nuget packages.
+            DeleteDirectory(Path.Combine(packageInstallDirectory, "/runtimes"));
```
chrisfromwork: would it be more appropriate to drop runtimes into a .External/External folder so that they aren't seen as assets but the content can still be accessed? I don't know if there are any nuget packages that only contain runtimes directories, but not being able to view that content may regress some NuGet for Unity's users experiences.

jwittner: Potentially, but I find it very unlikely because it would currently function if the only place they had a dll was in only one architecture under the runtimes folder. Otherwise they'd run into the same conflicting dll issues.

chrisfromwork:
Sounds good, I am fairly new to NuGet and will happily defer to your two's perspectives. :)

PR #242, opened 2019-04-09
fixed bug for line 324 of NugetHelper.cs; 

tk009999:
The reason is "The DotNetVersion will be 0 because if the InitializeOnLoad not loaded again"


Assets/NuGet/Editor/NugetHelper.cs
```diff
@@ -335,82 +348,82 @@ private static void Clean(NugetPackageIdentifier package)
 
                     // Select the highest .NET library available that is supported
                     // See here: https://docs.nuget.org/ndocs/schema/target-frameworks
-                    if (usingStandard2 && directoryName == "netstandard2.0")
+                    if (DotNetVersion == ApiCompatibilityLevel.NET_Standard_2_0 && directoryName == "netstandard2.0")
```
jwittner: In the comment above we were using the hardcoded values to support older versions of Unity. Are you sure this hasn't broken 5.3?

Assets/NuGet/Editor/NugetHelper.cs
```diff
@@ -441,6 +454,13 @@ private static void Clean(NugetPackageIdentifier package)
                         selectedDirectories.Add(directory.FullName);
                         break;
                     }
+                    else
+                    {
+                        string versionName = GetNuGetVersionName(directoryName);
+                        // Current API version is NET_Standard_2_0; NET_Standard_4_5 is required to support this NuGet package.
+                        Debug.LogError(string.Format("Current API version is {0}; {1} is required to support this NuGet package.", DotNetVersion, versionName));
```
jwittner: We're actually iterating over all the libraries trying to find a viable one - reporting an error here is too early I believe.

Assets/NuGet/Editor/NugetHelper.cs
```diff
@@ -1169,8 +1277,11 @@ public static bool Install(NugetPackage package, bool refreshAssets = true)
                 {
                     string baseDirectory = Path.Combine(NugetConfigFile.RepositoryPath, string.Format("{0}.{1}", package.Id, package.Version));
 
+
+                    ReadOptions options = new ReadOptions();
```
jwittner: Were you hitting errors and needed this? Is the encoding of nuget packages documented as always being utf8?

Assets/NuGet/Editor/NugetHelper.cs
```diff
@@ -1383,7 +1498,8 @@ public static Texture2D DownloadImage(string url)
             stopwatch.Start();
 
             bool fromCache = false;
-            if (ExistsInDiskCache(url)) {
+            if (ExistsInDiskCache(url))
```
jwittner: Please don't apply sweeping formatting changes and adhere to the formatting in place.

Assets/packages.config
```diff
@@ -1,4 +1,2 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
-<packages>
-  <package id="Newtonsoft.Json" version="6.0.8" />
-</packages>
\ No newline at end of file
+<packages />
```
jwittner: Do you have a strong reason to remove this development dependency?

ProjectSettings/ProjectVersion.txt
```diff
@@ -1 +1 @@
-m_EditorVersion: 2017.1.1f1
+m_EditorVersion: 2018.3.0f2
```
jwittner: While I appreciate a version upgrade as much as the next person, I'd like to avoid updating to versions that haven't had the security patch applied for https://unity3d.com/security?utm_campaign=engine_global_service_190301-gbl-security-rce&utm_content=190301-gbl-security-rce&utm_medium=email&utm_source=Eloqua#vulnerability-details

jwittner: Did you have a strong need for the version update?

Assets/NuGet/Editor/NugetHelper.cs
```diff
@@ -94,9 +94,10 @@ static NugetHelper()
             {
                 return;
             }
-            
+
 #if UNITY_5_6_OR_NEWER
             DotNetVersion = PlayerSettings.GetApiCompatibilityLevel(EditorUserBuildSettings.selectedBuildTargetGroup);
+            Debug.LogWarning(string.Format("The project's API version is {0} attention please the API version supproted you want to download the nuget package.", DotNetVersion));
```
jwittner: I'm not sure why we'd always want a warning for older versions of Unity?

tk009999: Thanks for noticing me. It's fixed already.

Assets/NuGet/Editor/NugetHelper.cs
tk009999: Sorry about that. I didn't follow the formatting and I fixed that already.

Assets/NuGet/Editor/NugetHelper.cs
tk009999: that's options was removed already.

Assets/NuGet/Editor/NugetHelper.cs
tk009999: that's LogError was removed already.

Assets/NuGet/Editor/NugetHelper.cs
tk009999: It's fixed changes to the original.

ProjectSettings/ProjectVersion.txt
tk009999: fixed all changes to the original.

Assets/NuGet/Editor/NugetHelper.cs
```diff
@@ -441,6 +447,13 @@ private static void Clean(NugetPackageIdentifier package)
                         selectedDirectories.Add(directory.FullName);
                         break;
                     }
+                    else
+                    {
+                        string versionName = GetNuGetVersionName(directoryName);
+                        // Current API version is NET_Standard_2_0; NET_Standard_4_5 is required to support this NuGet package.
+                        Debug.LogWarning(string.Format("Current API version is {0}; {1} is required to support this NuGet package.", DotNetVersion, versionName));
```
jwittner: I think a warning is too strong here as it's expected that there would be other target frameworks that don't work with Unity and this code is determining which of all the options best fits the current Unity settings. NuGetForUnity has a verbose logging concept that I think is a better fit for this point in the code. [Here's where](https://github.com/GlitchEnzo/NuGetForUnity/blob/911f02a5dc4d730139478f1805dc6d2990c5e3af/Assets/NuGet/Editor/NugetHelper.cs#L449) we log what folder we're selecting.

Maybe what you're looking for is a warning (or maybe even an error) if there are available target frameworks, but none of them can work with Unity's current settings? Testing `libDirectories.Count > 0 && selectedDIrectories.Count == 0` [here](https://github.com/GlitchEnzo/NuGetForUnity/blob/911f02a5dc4d730139478f1805dc6d2990c5e3af/Assets/NuGet/Editor/NugetHelper.cs#L445) would give you that.

Assets/NuGet/Editor/NugetHelper.cs
```diff
@@ -583,6 +596,94 @@ private static void Clean(NugetPackageIdentifier package)
             }
         }
 
+        private static string GetNuGetVersionName(string directoryName)
+        {
+            string versionName = "";
+
+            if (directoryName == "netstandard2.0")
```
jwittner: I'm hoping we can avoid this whole mapping (including the need to maintain it), rephrase the log slightly to leverage the directory names themselves and point users here: https://docs.microsoft.com/en-us/nuget/reference/target-frameworks?

jwittner: And potentially here https://docs.unity3d.com/ScriptReference/ApiCompatibilityLevel.html

tk009999: OK. I got your mean. I fixed the redundancy judgment.

Assets/NuGet/Editor/NugetHelper.cs
tk009999: Sure, this recommended is good. I fixed it already. ;D

Assets/NuGet/Editor/NugetHelper.cs
jwittner: Can you delete this method now that no one calls it?

jwittner: Can you remove this function now that it's unused?

Assets/NuGet/Editor/NugetHelper.cs
```diff
@@ -442,6 +448,16 @@ private static void Clean(NugetPackageIdentifier package)
...
+                
+                // Checks if Api Compatibility Level in PlayerSettings matches the version requried for this NuGet package.                
+                if (libDirectories.Count() > 0 && selectedDirectories.Count == 0) {
+                    foreach (var directory in libDirectories)
+                    {
+                        string directoryName = directory.Name.ToLower();
+                        // Current API version is NET_Standard_2_0; NET_Standard_4_5 is required to support this NuGet package.
+                        LogVerbose("Current API version is {0}; {1} is required to support this NuGet package.", DotNetVersion, directoryName);
+                    }                                            
+                }
```
jwittner: While I think this might be an error case that should prevent the package from installing, theres no way to shortcut that from this function right now so, I'm hoping you won't mind logging a warning right here about the general issue? Something like "{PackageName} has lib folders, but none supported by the current API Version {DotNetVersion}"?

Please keep the verbose prints, but also log a warning outside of your foreach about the failure to find a compatible lib.

jwittner:
I believe recent changes have superseded this PR.

PR #244, opened 2019-04-22
Support new ado packge url shape

### LINKED ISSUE #236: Update hostname to use with CredentialProvider
## Description
Some time ago, Visual Studio Team Services (VSTS) was renamed to Azure DevOps. With that name change the URLs were also changed. That's why the code in [NugetHelper.cs#L1250](https://github.com/GlitchEnzo/NuGetForUnity/blob/master/Assets/NuGet/Editor/NugetHelper.cs#L1250) now doesn't work any more.

The new URL format for package feeds is `https://pkgs.dev.azure.com/[organization name]/_packaging/[feed name]/nuget/v3/index.json`

It still works fine if you use the old URL (which is automatically redirected to the new URLs), so NuGetForUnity should support both.

Note that the download URL of the CredentialProviderBundle has to be `https://pkgs.dev.azure.com/[organization name]/_apis/public/nuget/client/CredentialProviderBundle.zip` so just passing the host name to `GetPasswordFromVSTSCredentialProvider` won't be enough any more (the organization name was a part of the host name before, now it isn't).

Cc @CoPrez because he implemented that in #164

- **NuGet Package:** -
- **NuGetForUnity Version:** 1.0.3
- **Unity Version:** 2018.3.5

jwittner:
Resolves #236 by adding support for the shape of the new package feed urls.

### DIFF
--- a/Assets/NuGet/Editor/NugetHelper.cs
+++ b/Assets/NuGet/Editor/NugetHelper.cs
@@ -1156,12 +1156,9 @@
                             EditorUtility.DisplayProgressBar(string.Format("Installing {0} {1}", package.Id, package.Version), "Downloading Package", 0.3f);
 
                         Stream objStream = RequestUrl(package.DownloadUrl, package.PackageSource.UserName, package.PackageSource.ExpandedPassword, timeOut: null);
-                        if (objStream != null)
+                        using (Stream file = File.Create(cachedPackagePath))
                         {
-                            using (Stream file = File.Create(cachedPackagePath))
-                            {
-                                CopyStream(objStream, file);
-                            }
+                            CopyStream(objStream, file);
                         }
                     }
                 }
--- a/Assets/NuGet/Editor/NugetHelper.cs
+++ b/Assets/NuGet/Editor/NugetHelper.cs
@@ -12,6 +12,7 @@
     using UnityEngine;
     using Debug = UnityEngine.Debug;
     using System.Security.Cryptography;
+    using System.Text.RegularExpressions;
 
     /// <summary>
     /// A set of helper methods that act as a wrapper around nuget.exe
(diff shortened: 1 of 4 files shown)


Assets/NuGet/Editor/NugetHelper.cs
```diff
@@ -1251,11 +1251,18 @@ public static Stream RequestUrl(string url, string password, int? timeOut)
...
 
-            if (packageHost.EndsWith("pkgs.visualstudio.com") && string.IsNullOrEmpty(password))
+            if (string.IsNullOrEmpty(password))
             {
-                // The host is a VisualStudio feed (which requires authentication) but a password was not provided. Use the VSS credential provider to aquire a token and append
-                // it to the request.
-                password = GetPasswordFromVSTSCredentialProvider(packageHost);
+                bool isKnownAuthenticatedFeed =
+                    packageHost.EndsWith("pkgs.visualstudio.com") ||
+                    packageHost.EndsWith("pkgs.dev.azure.com");
```
andreiborodin: Could we just put a checkbox on the feed in the settings to say "Use Credential Manager"?

It's a suggestion, not a blocker :)

jwittner: Really like this suggestion, added it as an enhancement #245 . I think I'll hold off on biting that off right now though. =)

Assets/NuGet/Editor/NugetHelper.cs
```diff
@@ -1154,10 +1155,13 @@ public static bool Install(NugetPackage package, bool refreshAssets = true)
                         if (refreshAssets)
                             EditorUtility.DisplayProgressBar(string.Format("Installing {0} {1}", package.Id, package.Version), "Downloading Package", 0.3f);
 
-                        Stream objStream = RequestUrl(package.DownloadUrl, package.PackageSource.ExpandedPassword, timeOut: null);
-                        using (Stream file = File.Create(cachedPackagePath))
+                        Stream objStream = RequestUrl(package.DownloadUrl, package.PackageSource.UserName, package.PackageSource.ExpandedPassword, timeOut: null);
+                        if (objStream != null)
```
andreiborodin: >  if (objStream != null) [](start = 22, length = 24)

Is this expected, or an error? Maybe print like a warning?

jwittner: Oops, a remnant from a different attempted implementation of RequestUrl.

PR #251, opened 2019-05-30
speed up refresh by issuing multiple GetUpdatesFallback requests in parallel

### LINKED ISSUE #250: speed up refresh by issuing multiple requests in parallel
## Description

On our team it typically takes 30-45 seconds to bring up the NuGet window or to click the Refresh button. The amount of time grows with the number of feeds and packages installed.

# Details

Inside NugetWindow:Refresh, web requests are issued one at a time (serially). It should be possible to query all feeds and updates simultaneously. For my team, a lot of the time spent is inside NuGetPackageSource.GetUpdates. GetUpdatesFallback must be used because Azure DevOps doesn't support the GetUpdates API for querying multiple packages in a single call. Each installed package creates a separate web call inside GetUpdatesFallback that takes about 4 seconds, adding up for projects that use a lot of different packages.

- **Unity Version:** 2017.4.22f1, 2018.3.9f1

# Suggestion

Run multiple web requests in parallel. This could include fetching from multiple feeds and fetching package updates. I would recommend waiting until they have all completed before displaying results in the UI because partial refreshes seem like they would be confusing and error-prone. I'm not sure of the best way to implement this. C# async/await might be the simplest way to organize this. Other options include [editor coroutines](https://docs.unity3d.com/Packages/com.unity.editorcoroutines@0.0/manual/index.html) and background threads.

Note that #6 may be a related issue/implementation. #6 would help unblock the Unity UI while waiting. However, we would still have to wait before the updated NuGetForUnity information is available unless requests are issued in parallel.

ForrestTrepte:
Resolves #250 

use BeginInvoke to run GetUpdatesFallback on a background thread

Also includes changes to classes used by GetUpdatesFallback to avoid Unity functions that aren't allowed on a background thread.

### DIFF
--- a/Assets/NuGet/Editor/NugetHelper.cs
+++ b/Assets/NuGet/Editor/NugetHelper.cs
@@ -95,6 +95,11 @@
         private static string applicationDataPath = Application.dataPath;
 
         /// <summary>
+        /// Cached value of Application.temporaryCachePath, available for use from a background thread.
+        /// </summary>
+        private static string applicationTemporaryCachePath = Application.temporaryCachePath;
+
+        /// <summary>
         /// The current .NET version being used (2.0 [actually 3.5], 4.6, etc).
         /// </summary>
         internal static ApiCompatibilityLevel DotNetVersion;
@@ -1502,7 +1507,7 @@
 
         private static string GetFilePath(string url)
         {
-            return Path.Combine(Application.temporaryCachePath, GetHash(url));
+            return Path.Combine(applicationTemporaryCachePath, GetHash(url));
         }
 
         private static string GetHash(string s)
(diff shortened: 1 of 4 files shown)


ForrestTrepte:
I'm of two minds about this change:
1. It does speed up the refresh operation, which is good. For a project with 4-5 packages, I'm seeing the time used to load updates decrease from ~20 seconds to ~5 seconds.
1. This makes the codebase significantly more difficult to understand, which is bad. This has the potential to introduce bugs due to multithreading timing corner cases or Unity functionality that doesn't play nice on a background thread. This code will be more difficult to maintain.

Not sure if it's worth it. I'm open to alternative ideas. Maybe it would be better to try a solution to #6 (put all of the refresh on a background thread) while leaving it to (more slowly) issue its requests serially? Or maybe we should require .NET 4.6 and use async/await? Or maybe this whole issue should be abandoned for the sake of simplicity?

ForrestTrepte:
It is looking like #254 will speed things up so that this may be unnecessary.

ForrestTrepte:
I'm going to abandon this. It turned out that fetching credentials was taking up the bulk of the time. With cached credentials from #254, GetUpdatesFallback is now taking ~1 second in my tests so parallelizing it is no longer as important.

PR #254, opened 2019-06-05
cache calls to GetCredentialProvider

### LINKED ISSUE #253: refresh makes many redundant time-consuming calls to credential provider
## Description

RequestUrl is called for just about every operation that NuGetForUnity uses to retrieve data. It, in turn calls GetCredentialFromProvider, which takes several seconds to spawn a credentialprovider*.exe process. This happens redundantly for multiple similar calls. We should cache these credentials so we only need to do this once for each feed.

- **Unity Version:** 2017.4.22f1

ForrestTrepte:
resolves #253


ForrestTrepte:
@CoPrez, FYI

Issue #269, opened 2019-08-16
/unityplugin folder hierarchy is lost; only files in root are copied


dfields-msft:
## /unityplugin folder hierarchy is lost; only files in root are copied
Native plugins typically need to deliver different flavors for different architectures, which can't co-exist in the same folder since the files have the same names. Instead, they're typically placed in subfolders by architecture/platform, but the N4U handling of /unityplugin targets only copies files in the root folder, and discards subfolders. Instead, the whole unityplugin folder should be copied recursively to Assets/Plugins.

dfields-msft:
@jwittner FYI

jwittner:
Hey @dfields-msft, doing some research on the need for these folders. It looks like the plugins folder [is no longer necessary](https://docs.unity3d.com/Manual/UpgradeGuide5-Plugins.html) and hasn't been since Unity 5.x.

> "Plugins are no longer required to be placed in Assets\Plugins\<Platform\> folders"

Doing some tests using OpenCV I'm seeing this to be true in practice.

PR #277, opened 2019-09-10
Issue #269: Recursively copy /unityplugin folder to keep folder hierarchy and contents

julianparismorgan:
This fixes issue #269, where the /unityplugin folder hierarchy was previously lost and only files in root were copied.


jwittner:
Good change, and worked great. Verified build passed locally.

Issue #238, opened 2019-03-20
Importing package from local directory causes issues with nuget restore in visual studio


chrisfromwork:
## Description
I'm not sure what the expected behavior is for nuget for unity, but i have the following scenario:

1) I have a custom nuget package that drops binaries into a Plugins\WSA\x86 folder when imported via nuget for unity
2) I then create a uwp visual studio solution for this unity project
3) I then open the visual studio solution to build the project

Expected:
When I build my visual studio solution, I don't see any errors related to the imported nuget package. The dll's that i have imported to plugins folder are reused and resolve correctly for my application.

Observed:
When I build my visual studio solution, I see an error that NuGet Package restore failed for my project (likely because visual studio doesn't know about my local package). The application builds fine and works using the dlls that were imported into the plugins directory.

What is the expected behavior here? It appears that Nuget for unity creates a packages.config file within my unity assets folder that is then used in my visual studio solution. Should the packages.config actually be dropped into something that isn't observed by unity like an .External directory? Is there anything that can be done so that visual studio uses paths that have been declared in the Nuget for Unity settings?

- **NuGet Package:** Custom/local package
- **NuGetForUnity Version:** 1.0.3
- **Unity Version:** 2018.3.8f1
- **Operating System:** Windows 10

Issue #280, opened 2019-09-11
NuGet restore errors in VS when building IL2CPP


wiwei:
## Description
When building a Window Store App from Unity (IL2CPP), building and deploying in VS works fine, but there are errors that show in the error window that look like this:

Severity | Code | Description
-- | -- | --
Error |   | NuGet Package restore failed for project   TestNuGet: Unable to find version '2.0.1-20190910v2' of package   'Microsoft.MixedReality.Toolkit.Examples'.  C:\Program Files   (x86)\Microsoft SDKs\NuGetPackages\: Package   'Microsoft.MixedReality.Toolkit.Examples.2.0.1-20190910v2' is not found on   source 'C:\Program Files (x86)\Microsoft SDKs\NuGetPackages\'.  https://api.nuget.org/v3/index.json:   Package 'Microsoft.MixedReality.Toolkit.Examples.2.0.1-20190910v2' is not   found on source 'https://api.nuget.org/v3/index.json'.. Please see Error List   window for detailed warnings and errors.
Error |   | NuGet Package restore failed for project   TestNuGet: Unable to find version '2.0.1-20190910v2' of package   'Microsoft.MixedReality.Toolkit.Extensions'.  C:\Program Files   (x86)\Microsoft SDKs\NuGetPackages\: Package   'Microsoft.MixedReality.Toolkit.Extensions.2.0.1-20190910v2' is not found on   source 'C:\Program Files (x86)\Microsoft SDKs\NuGetPackages\'.  https://api.nuget.org/v3/index.json:   Package 'Microsoft.MixedReality.Toolkit.Extensions.2.0.1-20190910v2' is not   found on source 'https://api.nuget.org/v3/index.json'.. Please see Error List   window for detailed warnings and errors.
[...]

PR #283, opened 2019-09-23
Prevent package.config from being included in WSA Build Output

andreiborodin:
This is the fix for #280 issue.

When the packages.config file is created or updated, this change will check for it's asset importer settings and disable it as an asset for WSA build. This will prevent it from being included in the output of the IL2CPP build and causing errors during NuGet restore.

The change also includes some of my auto-formatting rules applied.

### DIFF
--- a/Assets/NuGet/Editor/PackagesConfigFile.cs
+++ b/Assets/NuGet/Editor/PackagesConfigFile.cs
@@ -57,8 +57,10 @@
         /// <returns>A newly created <see cref="PackagesConfigFile"/>.</returns>
         public static PackagesConfigFile Load(string filepath)
         {
-            PackagesConfigFile configFile = new PackagesConfigFile();
-            configFile.Packages = new List<NugetPackageIdentifier>();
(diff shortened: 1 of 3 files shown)


andreiborodin:
This also fixes #238 .

jwittner:
Testing locally.

Assets/NuGet/Editor/PackagesConfigFile.cs
```diff
@@ -125,6 +142,29 @@ public void Save(string filepath)
...
+        }
+
+        private static void CheckImportSettings(string filePath, bool notifyOfUpdate)
+        {
+            filePath = Path.GetFullPath(filePath);
+            PluginImporter importer = AssetImporter.GetAtPath(filePath.Replace(Path.GetFullPath(Application.dataPath), "Assets")) as PluginImporter;
+
+            if (importer == null)
+            {
+                Debug.LogError(string.Format("Couldn't get importer for '{0}'.", filePath));
```
jwittner: I hit this error when opening N4U from a fresh repo with this change.

Issue #178, opened 2018-08-13
An assembly `System.Net.Http` with the same identity has already been imported.


wingyplus:
## Description

I found an error when try to install `Google.Cloud.PubSub`. The stacktrace is:

```

Assets/Packages/System.Net.Http.4.1.1/ref/netstandard1.1/System.Net.Http.dll (Location of the symbol related to previous error)
Assets/Packages/System.Net.Http.4.1.1/runtimes/win/lib/netcore50/System.Net.Http.dll (Location of the symbol related to previous error)
error CS1703: An assembly `System.Net.Http' with the same identity has already been imported. Consider removing one of the references
Assets/Packages/System.Net.Http.4.1.1/ref/netstandard1.1/System.Net.Http.dll (Location of the symbol related to previous error)
Assets/Packages/System.Net.Http.4.1.1/runtimes/win/lib/netstandard1.3/System.Net.Http.dll (Location of the symbol related to previous error)
error CS1703: An assembly `System.Net.Http' with the same identity has already been imported. Consider removing one of the references
Assets/Packages/System.Net.Http.4.1.1/ref/netstandard1.1/System.Net.Http.dll (Location of the symbol related to previous error)
Assets/Packages/System.Net.Http.4.1.1/runtimes/win/lib/net46/System.Net.Http.dll (Location of the symbol related to previous error)
error CS1703: An assembly `System.Net.Http' with the same identity has already been imported. Consider removing one of the references
Assets/Packages/System.Net.Http.4.1.1/ref/netstandard1.1/System.Net.Http.dll (Location of the symbol related to previous error)
Assets/Packages/System.Net.Http.4.1.1/runtimes/unix/lib/netstandard1.6/System.Net.Http.dll (Location of the symbol related to previous error)
error CS1703: An assembly `System.Net.Http' with the same identity has already been imported. Consider removing one of the references
Assets/Packages/System.Net.Http.4.1.1/ref/netstandard1.1/System.Net.Http.dll (Location of the symbol related to previous error)
Assets/Packages/System.Net.Http.4.1.1/lib/net46/System.Net.Http.dll (Location of the symbol related to previous error)
```

- **NuGet Package: Google.Cloud.PubSub** [What is the name and version of the NuGet package you are installing?]
- **NuGetForUnity Version: 0.0.16** [What version of NuGetForUnity are you running?]
- **Unity Version: 2018.2.2f1** [What version of Unity are you using?]
[...]

PR #287, opened 2019-10-02
Delete ref folder as part of package cleanup

majaeger:
Fixes the issue described in: https://github.com/GlitchEnzo/NuGetForUnity/issues/178

The root cause is that the ref folder contains stub dlls for various platforms that are used as a compiler reference in Visual Studio. In Unity, we just need the implementation dll from the lib folder.

Testing: Verified that after the change, I was able to install a package that included a dependency on the System.Numerics.Vectors package with no errors.


majaeger:
The appveyor build appears to be broken unrelated to this change.

majaeger:
Thanks @jwittner! Do you have an ETA on the next time an updated release build will be published?

PR #295, opened 2019-10-09
added trim to preferences path to remove leading or trailing whitespaces

### LINKED ISSUE #249: Sanitize/Validate Source URLs
I mistakenly copy&pasted a source URL with an additional space at the end - this made requests fail. It wasn't difficult to figure out what happened, but it'd be great if those trailing spaces had been removed automatically, or a visible error message/dialog warned users when the URL didn't finish with a "/" character.

thalbern:
fixes #249 where nuget package source urls couldn't be opened when there was a whitespace in the end of the url

### DIFF
--- a/Assets/NuGet/Editor/NugetPreferences.cs
+++ b/Assets/NuGet/Editor/NugetPreferences.cs
@@ -61,7 +61,7 @@
                         EditorGUILayout.BeginVertical();
                         {
                             source.Name = EditorGUILayout.TextField(source.Name);
-                            source.SavedPath = EditorGUILayout.TextField(source.SavedPath);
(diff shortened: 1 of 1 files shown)


Assets/NuGet/Editor/NugetPreferences.cs
```diff
@@ -61,7 +61,7 @@ public static void PreferencesGUI()
                         EditorGUILayout.BeginVertical();
                         {
                             source.Name = EditorGUILayout.TextField(source.Name);
-                            source.SavedPath = EditorGUILayout.TextField(source.SavedPath);
+                            source.SavedPath = EditorGUILayout.TextField(source.SavedPath).Trim();
```
andreiborodin: Could this be null?

PR #314, opened 2019-11-27
Fix prerelease-vs-final version comparison

### LINKED ISSUE #313: Prerelease/final versions not compared correctly
## Description
Upgrading from a prelease version (e.g. 1.0.0-rc1) to the corresponding final version (e.g. 1.0.0) incorrectly displays a 'package downgrade' warning.

According to the [NuGet package versioning](https://docs.microsoft.com/en-us/nuget/concepts/package-versioning) spec, the meaning of a prerelease "-Suffix" is as per [Semantic Versioning 1.0](https://semver.org/spec/v1.0.0.html) conventions. Those say that "Pre-release versions satisfy but have a lower precedence than the associated normal version", and by way of an example: "1.0.0-rc1 < 1.0.0"

- **NuGet Package:** e.g. Microsoft.MixedReality.WebRTC.Native.UWP
- **NuGetForUnity Version:** 2.0.0
- **Unity Version:** 2018.4.6f1
- **Operating System:** Windows

michael-buschbeck-ms:
Fixes #313

### DIFF
--- a/Assets/NuGet/Editor/NugetPackageIdentifier.cs
+++ b/Assets/NuGet/Editor/NugetPackageIdentifier.cs
@@ -335,7 +335,7 @@
             {
                 string[] splitStringsA = versionA.Split('-');
                 versionA = splitStringsA[0];
-                string prereleaseA = string.Empty;
+                string prereleaseA = "\uFFFF";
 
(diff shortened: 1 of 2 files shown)


jwittner:
Works great!

PR #317, opened 2019-12-11
Usability: Added a "Preferences" button to "Manage Nuget Packages" window.

### LINKED ISSUE #232: Add 'Preferences -> NuGet For Unity' button to 'Manage NuGet Packages' UI
## Description
Based on this projects landing page, I did not know that the ability existed to set a custom path for locating nuget packages. I would find it helpful if a button was added to the NuGet -> Manage NuGet Packages window to launch the Nuget For Unity preferences. Otherwise, this window just appears to only support nuget.org for obtaining nuget packages. The 'Online' tab also didn't provide confidence that I would be able to install nuget packages located on my local machine compared to a server.

UnquotidianVi:
Added a "Preferences" button next to the "Refresh" button in "Manage Nuget Packages" window. When the button is pressed, it opens "Preferences/NuGet For Unity" window and closes the "Manage Nuget Packages" window. I decided to close the Manage window because most of the Preference options require the Manage window to be refreshed.

Fixes #232

### DIFF
--- a/Assets/NuGet/Editor/NugetWindow.cs
+++ b/Assets/NuGet/Editor/NugetWindow.cs
@@ -652,6 +652,16 @@ namespace NugetForUnity
                     {
                         Refresh(true);
                     }
+
+                    if (GUILayout.Button("Preferences", GUILayout.Width(80)))
+                    {
+#if UNITY_2018_3_OR_NEWER
+                        SettingsService.OpenUserPreferences("Preferences/NuGet For Unity");
+#else
+                        EditorApplication.ExecuteMenuItem("Edit/Preferences...");
+#endif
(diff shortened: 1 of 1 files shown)


Efmi:
`SettingsService.OpenUserPreferences()` is only supported in 2018.3 or never. We can still use `EditorApplication.ExecuteMenuItem("Edit/Preferences...");` on older versions, but this doesn't allow us to open the specific Preference menus.

Efmi:
I hadn't realized that the Preferences button only showed up on the 'Online' -tab. Everything should be ready for review now. 👍

JoC0de:
I rebased ad reformatted your code. It works 👍 
So thank you @UnquotidianVi for the improvement.

PR #365, opened 2021-01-15
Upgrade projects to Unity 2018.4 / .Net 4.7.1

jwittner:
These changes deprecate support for versions of Unity prior to 2018.4, which Unity no longer supports itself. It also fixes builds for some of the recent changes don't compile in .Net 3.x.


build.ps1
```diff
@@ -41,10 +41,10 @@ Copy-Item ".\CreateDLL\bin\Release\NugetForUnity.dll" ".\Packager\Assets\NuGet\E
 Copy-Item ".\CreateDLL\bin\Release\DotNetZip.dll" ".\Packager\Assets\NuGet\Editor"
 
 # Launch Unity to export the NuGetForUnity package
-Start-UnityEditor -Project ".\Packager" -BatchMode -Quit -Wait -ExportPackage "Assets/NuGet .\NuGetForUnity.unitypackage" -LogFile ".\Packager\NuGetForUnity.unitypackage.log"
+Start-UnityEditor -Project ".\Packager" -BatchMode -Quit -Wait -ExecuteMethod "NugetForUnity.Export.Execute" -LogFile ".\Packager\NuGetForUnity.unitypackage.log"
```
jwittner, line 44: For an unknown reason, exporting packages via -ExportPackage do not work for me. To ensure this doesn't happen to others I've moved to using a script method which does the same thing.

jwittner:
@KonH - these updates are necessary to get a release out for your changes (some APIs you used don't exist in the older versions that we were supporting). If you can verify nothing seems broken for you I can merge this in, make a build, and push a release for v3.0.0.

KonH:
@jwittner these changes work as expected on a few test cases. I think it's fine to merge them.

PR #368, opened 2021-01-28
Drop DotNetZip

### LINKED ISSUE #261: Zip conflicts
## Description

Installing NuGetForUnity creates conflicts with previously installed versions of DotNetZip.

Seems to only affect scripts compiled for the editor.

Workaround is to disable existing DotNetZip for editor. But this is not ideal: the version of DotNetZip won't be the same for scripts running in the editor versus built player.

Is there a way to disable the bundled DotNetZip that comes with NuGetForUnity? Delete it completely? I might try that.

What would happen if I tried to install DotNetZip through NuGetForUnity?

- **NuGet Package:** n/a
- **NuGetForUnity Version:** 0.0.13
- **Unity Version:** 2018.4.3.1f
- **Operating System:** macOS Mojave 10.14.5

### LINKED ISSUE #282: Ionic.Zip.ZipException: Cannot read that as a ZipFile
## Description
Upon starting up Unity, the console reports `Ionic.Zip.ZipException: Cannot read that as a ZipFile`. When attempting to install AssimpNet, the console reports `BadReadException: Could not read block - no data!  (position 0x0006BA4F)`. The package does not install.

- **NuGet Package:** None, AssimpNet
- **NuGetForUnity Version:** 0.0.13
- **Unity Version:** 2019.2.3f1
- **Operating System:** Windows 10 64bit, build 18362

jwittner:
New updates have moved us to Unity 2018.4, which put leveraging System.IO.Compression back on the table. This update moves from DotNetZip to just that.

Fixes #261
No repro for this issue, but I only tested with these changes, so fixes #282


GlitchEnzo:
Awesome!

derekfreed (approved):

Issue #383, opened 2021-03-24
Could not install package because of Nuget V2 deprecation


sajjadrad:
## Description
I got api deprecation error while installing a package

- **NuGet Package:** System.Buffers [4.4.0, ) or greater
- **NuGetForUnity Version:** 3.0.1
- **Unity Version:** 2020.1.3f1
- **Operating System:** macOS 11.2

```Unable to retrieve package list from http://www.nuget.org/api/v2/FindPackagesById()?id='System.Buffers'&$orderby=Version asc
System.Net.WebException: The remote server returned an error: (400) Bad Request.
  at System.Net.HttpWebRequest+<GetResponseFromData>d__240.MoveNext () [0x00152] in <525dc68fbe6640f483d9939a51075a29>:0 
--- End of stack trace from previous location where exception was thrown ---
  at System.Net.HttpWebRequest.GetResponse () [0x00013] in <525dc68fbe6640f483d9939a51075a29>:0 
  at NugetForUnity.NugetHelper.RequestUrl (System.String url, System.String userName, System.String password, System.Nullable 1[T] timeOut) [0x000b9] in <ebb7710be32040ba9a0cb144ce2a8967>:0 
  at NugetForUnity.NugetPackageSource.GetPackagesFromUrl (System.String url, System.String username, System.String password) [0x0006c] in <ebb7710be32040ba9a0cb144ce2a8967>:0 
  at NugetForUnity.NugetPackageSource.FindPackagesById (NugetForUnity.NugetPackageIdentifier package) [0x000ee] in <ebb7710be32040ba9a0cb144ce2a8967>:0 
UnityEngine.Debug:LogErrorFormat(String, Object[])
NugetForUnity.NugetPackageSource:FindPackagesById(NugetPackageIdentifier)
NugetForUnity.NugetPackageSource:GetSpecificPackage(NugetPackageIdentifier)
NugetForUnity.NugetHelper:GetOnlinePackage(NugetPackageIdentifier)
NugetForUnity.NugetHelper:GetSpecificPackage(NugetPackageIdentifier)
NugetForUnity.NugetHelper:InstallIdentifier(NugetPackageIdentifier, Boolean)
NugetForUnity.NugetHelper:Install(NugetPackage, Boolean)
NugetForUnity.NugetHelper:InstallIdentifier(NugetPackageIdentifier, Boolean)
NugetForUnity.NugetHelper:Install(NugetPackage, Boolean)
[...]

PR #384, opened 2021-03-26
Quick fix for V2 OData NuGet queries.

opexsdc:
Some folks have hit the issue #383 with the recent changes to NuGet V2 OData support.  V2 OData queries are still supported, IF you drop ```&$orderby=Version asc``` in the request.  This probably breaks other things, but at least you can carry on using this in Unity for now.

See: https://devblogs.microsoft.com/nuget/custom-v2-odata-queries-will-be-deprecated-march-9-2021/

### DIFF
--- a/Assets/NuGet/Editor/NugetPackageSource.cs
+++ b/Assets/NuGet/Editor/NugetPackageSource.cs
@@ -133,7 +133,6 @@
             else
             {
                 // See here: http://www.odata.org/documentation/odata-version-2-0/uri-conventions/
-                // Note: without $orderby=Version, the Version filter below will not work
                 string url = string.Format("{0}FindPackagesById()?id='{1}'", ExpandedPath, package.Id);
                 
                 // Are we looking for a specific package?
--- a/Assets/NuGet/Editor/NugetPackageSource.cs
+++ b/Assets/NuGet/Editor/NugetPackageSource.cs
@@ -134,7 +134,7 @@
             {
                 // See here: http://www.odata.org/documentation/odata-version-2-0/uri-conventions/
                 // Note: without $orderby=Version, the Version filter below will not work
-                string url = string.Format("{0}FindPackagesById()?id='{1}'&$orderby=Version asc", ExpandedPath, package.Id);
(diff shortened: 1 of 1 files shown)


anchpop:
Thanks! @jwittner any chance of getting this merged?

jwittner:
Thanks for the PR @opexsdc , and for the re-ping @anchpop . =) Can I get a good package config attached here that reproduces the issue? Want to be sure I'm testing the right problem with this PR. I'll pull this down tomorrow and give it a few run throughs with some of the examples from #383 as well.

Then I'll do a bit of smoke testing of the rest of the features to make sure this doesn't break anything else. =)

jwittner:
From perusing the code I think the issue may come from packages that don't have explicit version dependencies. This code should only be running now if there is a version range coming in. Worried that the proposed fix won't find a version correctly in the range being specified.

opexsdc:
Apologies - I should have junked the comment on line 136 above: ```// Note: without $orderby=Version, the Version filter below will not work``` no longer applies with the current operation of the NuGet V2 OData API.  The code is executing for non-local paths and the version filter applied below on line ~142 still works with this fix.

e.g.
Without version filter:
[https://www.nuget.org/api/v2/FindPackagesById()?id='System.Threading.Tasks.Extensions'](url)

With version filter:
[https://www.nuget.org/api/v2/FindPackagesById()?id='System.Threading.Tasks.Extensions'&$filter=Version%20eq%20'4.0.0-rc2-24027'](url)

----

I have attached a ```packages.config``` from where I used this Unity package - I'm not sure if it is any _good_ since I only needed one package + deps... 
[https://gist.github.com/opexsdc/80b29da4cd8b2224758e7dcdf0672c0b#file-packages-config](url)

jwittner:
Largely this works great! While testing the cases from the bug I ran into some exceptions about version ranges, which led me to also realize that we were making FindPackageById requests for packages we weren't going to install anyways, as they are already imported by the Unity engine.

I've made a PR to address the version range issues and the unecessary request issues: #385 

Would much appreciate any tests on that PR by those seeing the 404 this PR addresses - the code path is followed only for projects and packages that specify dependencies with version ranges, so you'll likely see some of the errors that my PR addresses even after taking this change. =)

PR #385, opened 2021-04-03
Improve version range handling, avoid unecessary package requests.

jwittner:
Improvements:

* Early out in more places for packages that are imported by the engine.
  * This avoids issuing queries for dependencies that we're not going to install anyways.
* Improves handling of version ranges.
  * Properly compares ranges in many cases where it was falling down.
  * Fixes several exceptions when comparing package versions.

### DIFF
--- a/Assets/NuGet/Editor/NugetHelper.cs
+++ b/Assets/NuGet/Editor/NugetHelper.cs
@@ -1606,12 +1606,13 @@
             string[] directories = Directory.GetDirectories(NugetConfigFile.RepositoryPath, "*", SearchOption.TopDirectoryOnly);
             foreach (string folder in directories)
             {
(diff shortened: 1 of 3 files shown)


DWishR:
I'll admit the absolute silence when attempting to install a package already included by Unity to be a little bit of an issue for me. Could up the severity of the log on Assets/NuGet/Editor/NugetHelper.cs ~1241. As it is, I can continually mash the install button for e.g. `System.Numerics.Vectors` and it happily ignores my requests. I think the best would be to grey out the "Install" button and have it say "Included" or some such.

jwittner:
That's great feedback @DWishR ! I was thinking that same thing as I noticed I only got some output when I had enabled Verbose Logging in the NugetForUnity preferences.

Issue #14, opened 2015-12-23
Support native libraries


GlitchEnzo:
Unity is terrible with it's handling of native DLLs, so this tool must provide some sort of additional support for them.

Unity requires native DLLs to be in the Assets/Plugins directory BUT Unity locks the files and there is no way to get it to release the lock without closing the Editor.  This creates complications such as saving any open scene, closing, copying/overwriting whatever DLLs need to be copied, re-opening Unity with whatever scene/project it had open, and continuing with the installation of the packages.

GlitchEnzo:
Since it is the Unity Editor itself that locks the files, it's doubtful that there is any change that can be done to NuGetForUnity to resolve this.

jwittner:
I don't believe the Unity plugin folder is a requirement anymore.

PR #393, opened 2021-05-04
Add Roslyn analyzer package support

nowsprinting:
Roslyn analyzers supported on Unity 2020.2 or later.

The analyzer needs to make the following settings in plugin import settings.
- Off all platforms
- Set label `RoslynAnalyzer`

With this setting, it will be enabled as an analyzer in Unity 2020.2 or later, and enabled in generated .csproj by the following plugins.
- JetBrains Rider Editor v2.0.6+
- Visual Studio Code Editor v1.2.0+

see more information
https://docs.unity3d.com/2020.2/Documentation/Manual/roslyn-analyzers.html


Assets/NuGet/Editor/NugetHelper.cs
```diff
@@ -1448,12 +1448,37 @@ public static bool Install(NugetPackage package, bool refreshAssets = true)
...
+                    ModifyImportSettingsIfRoslynAnalyzer(package.Id, package.Version);
                     EditorUtility.ClearProgressBar();
                 }
             }
             return installSuccess;
         }
 
+        private static void ModifyImportSettingsIfRoslynAnalyzer(string id, string version)
+        {
+            var dir = $"Assets/Packages/{id}.{version}";
```
robin-moss, line 1460: You should use `NugetConfigFile.RepositoryPath` instead of the hardcoded path

robin-moss, line 1460:
```suggestion
            var dir = Path.Combine(NugetConfigFile.RepositoryPath, $"{id}.{version}");
            if (!Directory.Exists(dir))
            {
                return;
            }
```

robin-moss: Ah but you do need relative paths :thinking:, could take the same approach as used when adding support for [native files](https://github.com/GlitchEnzo/NuGetForUnity/pull/421/files#diff-460028a9a166ac23a7e76cf1282330ca51138fc1d7557336de43c60183de8f44R1393)

robin-moss: To get it working locally I made the following changes

```c#
var dir = Path.Combine(NugetConfigFile.RepositoryPath, $"{id}.{version}");
if (!Directory.Exists(dir))
{
    return;
}
foreach (var analyzer in Directory.GetFiles(dir, "*.dll", SearchOption.AllDirectories)
    .Where(x => x.Contains("/analyzers/")))
{
    // Trigger Unity to create the meta file, the path for ImportAsset MUST be relative to the project
    // path
    var projectRelativePath =
        analyzer.Substring(new DirectoryInfo(Application.dataPath).Parent.FullName.Length + 1);
    AssetDatabase.ImportAsset(projectRelativePath, ImportAssetOptions.ForceSynchronousImport);
    var meta = AssetImporter.GetAtPath(projectRelativePath) as PluginImporter;
```

nowsprinting: Fixed 9d60da5

nowsprinting:
@robin-moss Thank you for your suggestions.

JoC0de:
I was able to improve the implementation using a combination of your (@nowsprinting) code and things @robin-moss done in #14 
So I am happy with the solution and how it works so I will merge it.
so @nowsprinting Thank you for the PR.
and @robin-moss Thank you for the "improvements".

PR #347, opened 2020-09-18
Added support for installing via UPM.

neogeek:
This PR adds support for installing this project via UPM (Unity Package Manager).

Read more at:
1. https://docs.unity3d.com/Manual/Packages.html
1. https://neogeek.dev/creating-custom-packages-for-unity-2018.3/

The changes made in this PR:
1. Added an assembly reference file in the main project.
1. Added an assembly reference in the test folder to limit access to the editor only.
1. Added a package.json file with information about the project. This file is used by Unity to identify the project as a UPM repo.
1. Changed access modifiers from internal to public to allow access to the main project files from the test folder.

To deploy changes from master to a branch that can be installed by UPM, push the latest changes to a subtree branch named `upm` using the following command:

```bash
$ git subtree push --prefix Assets origin upm
```

Once the changes are pushed, the project can be installed by adding https://github.com/GlitchEnzo/NuGetForUnity.git#upm via the UPM panel by selecting **Add packages from git URL.**

You can also deploy versions that can be installed via UPM by creating a new tag/release from the `upm` branch. Once released, versions can then be installed by adding the tag to the end of the URL in place of `upm`.

https://github.com/GlitchEnzo/NuGetForUnity.git#upm
https://github.com/GlitchEnzo/NuGetForUnity.git#v1.0.0

Let me know if you have any questions about this PR.

PR #394, opened 2021-05-08
UPM support based on #347

nowsprinting:
This PR based on #347

I'm sorry I couldn't fork because of another PR > @neogeek
Even if you discard this PR and cherry-pick it to #347

Fix:
- Required Unity version
- Fix test asmdef format for Unity 2019.2 or earlier

discussions:

> Thinking we should push just the NuGet folder as the root and not push the Tests folder at all.

Popularly, the Tests folder is also published.
see: https://docs.unity3d.com/Manual/cus-layout.html

And, I worried about the package name.
Does anyone own domain `glitchenzo.com` that corresponds to package name `com.glitchenzo`?
I think `com.github.glitchenzo` is safe.


JoC0de:
Implement it in #480. Used the `Readme.md` and the 'package name' from your branch (with slightly changes).
@nowsprinting Thanks for the contribution.

PR #413, opened 2021-08-27
Making the project compatible with Unity Package Manager

badjano:
- Making it UPM compatible
- new instructions


cyraid:
This would be pretty nice ^

badjano:
> Curious what you mean by compatible here? I've used this alongside UPM for quite some time now. This is a rather big change so I'm hoping to understand the intent before looking through it all.

the change is that you can add the git url directly into packagemanage, so all the files in this repository are not included on your project, meaning less storage needed on your repository servers

Thaina:
> Curious what you mean by compatible here? I've used this alongside UPM for quite some time now. This is a rather big change so I'm hoping to understand the intent before looking through it all.

I think you know we always want to use package manager like UPM and nuget, so why this package itself still being `unitypackage` and cannot be installed with UPM ?

badjano:
> > Curious what you mean by compatible here? I've used this alongside UPM for quite some time now. This is a rather big change so I'm hoping to understand the intent before looking through it all.
> 
> I think you know we always want to use package manager like UPM and nuget, so why this package itself still being `unitypackage` and cannot be installed with UPM ?

there can also be an even better solution, by installing nuget libraries on packages folder ( a bit hackish since they are not designed to be so, but it would work with a little file managing ).

badjano:
I know this pull request is a bit large, but it is basically moving folders and adding a json file for the package manager to understand the whole thing, but if you guys would rather continue with the .unitypackage version, I guess it´s ok, but I´ll try to keep my repository updated with anything commited here, and also maybe make some more changes so that less garbage is left on the assets folder

lufinkey:
@GlitchEnzo can you please address this? It would be really great to have a trusted UPM source for this package

JoC0de:
#394 
#347 
#436

JoC0de:
Implement it in #480. But I kept the `.unitypackage` installation method as fallback. 
@badjano Thanks for the inspiration.

PR #417, opened 2021-09-27
Add configurable web request timeouts

RichLogan:
I had some issues with installs timing out, so I needed to add a config option to set a custom timeout for all Nuget operations. I also removed the hardcoded defaults to always use the config file setting, which stays at the original default of 5000ms. Thought it made sense as a feature for others 👍🏻


JoC0de:
Is included in #518

PR #421, opened 2021-10-28
feat: Add support for native runtimes

robin-moss:
Making use of PluginImporter to allow us to configure the compatibility of native libraries to different platforms

### DIFF
--- a/Assets/PlayTests/NuGetPlayTests.cs
+++ b/Assets/PlayTests/NuGetPlayTests.cs
@@ -1,0 +1,76 @@
+using System;
+using System.Collections;
+using System.Runtime.InteropServices;
+using NugetForUnity;
+using NUnit.Framework;
+using UnityEngine;
+using UnityEngine.TestTools;
+
+[assembly: PrebuildSetup(typeof(NuGetPlayTests))]
+[assembly: PostBuildCleanup(typeof(NuGetPlayTests))]
+
+/// <summary>
+/// Play mode tests allow us to install NuGet packages with Native code before play mode starts, then when play mode
+/// runs the native libraries are available for use.
+///
+/// There seems to be some Unity internals that prevents an edit mode test from adding a Native library and finding it
+/// in the same run. 
+/// </summary>
+public class NuGetPlayTests : IPrebuildSetup, IPostBuildCleanup
+{
+    readonly NugetPackageIdentifier sqlite = new NugetPackageIdentifier("SQLitePCLRaw.lib.e_sqlite3", "2.0.7");
+
+    /// <summary>
+    /// This is the version number of sqlite with periods replaced with zeros.
+    ///
+    /// Note: Version of the SQLite library does not match the NuGet package version 
+    /// </summary>
+    private readonly int _expectedVersion = 3035005;
+
+    [UnityTest]
+    public IEnumerator InstallAndRunSqlite()
+    {
+        yield return new WaitForFixedUpdate();
+
+        // Test the actual library by calling a "extern" method
+        var version = sqlite3_libversion_number();
+        Assert.That(version, Is.EqualTo(_expectedVersion));
+    }
+
+    /// <summary>
+    /// Call to the SQLite native file, this actually tests loading and access the library when called.
+    ///
+    /// On windows the call to sqlite3_libversion returns a garbled string so use the integer
+    /// </summary>
+    /// <returns></returns>
+    [DllImport("e_sqlite3")]
+    private static extern int sqlite3_libversion_number();
+
+    public void Setup()
+    {
+        try
+        {
+            sqlite3_libversion_number();
+            Assert.Fail("e_sqlite3 dll loaded, but should not be available");
+        }
+        catch (DllNotFoundException)
+        {
+        }
+
+        // For windows we end up importing two identical DLLs temporarily, this causes an error log that NUnit detects
+        // and would fail the test if we don't tell it to ignore the Failing messages
+        NugetHelper.LoadNugetConfigFile();
+        NugetHelper.LoadSettingFile();
+        LogAssert.ignoreFailingMessages = true;
+        NugetHelper.InstallIdentifier(sqlite);
+        Assert.IsTrue(NugetHelper.IsInstalled(sqlite), "The package was NOT installed: {0} {1}", sqlite.Id,
+            sqlite.Version);
+    }
+
+    public void Cleanup()
+    {
+        NugetHelper.Uninstall(sqlite);
+        Assert.IsFalse(NugetHelper.IsInstalled(sqlite), "The packages are STILL installed: {0} {1}", sqlite.Id,
+            sqlite.Version);
+    }
+}
--- a/Assets/NuGet/Editor/MetaFileHandler.cs
+++ b/Assets/NuGet/Editor/MetaFileHandler.cs
@@ -1,0 +1,203 @@
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Reflection;
+using UnityEditor;
+using UnityEngine;
+
+namespace NugetForUnity
+{
+    public class MetaFileHandler : AssetPostprocessor
+    {
+        private const string RuntimesFolderName = "runtimes";
+        private const string NativeFolderName = "native";
+        private const string AssetsFolderName = "Assets";
+
+        /// <summary>
+        /// Used to mark an asset as already processed by this class.
+        /// </summary>
+        private const string ProcessedLabel = "NuGetForUnity";
+
+        private static readonly List<BuildTarget> NotObsoleteBuildTargets = typeof(BuildTarget)
+            .GetFields(BindingFlags.Public | BindingFlags.Static)
+            .Where(fieldInfo => fieldInfo.GetCustomAttribute(typeof(ObsoleteAttribute)) == null)
+            .Select(fieldInfo => (BuildTarget)fieldInfo.GetValue(null))
+            .ToList();
+
+        private enum ProcessState
+        {
+            Success,
+            Failure,
+            AlreadyProcessed
+        }
+
+        private static void OnPostprocessAllAssets(string[] importedAssets, string[] deletedAssets,
+            string[] movedAssets,
+            string[] movedFromAssetPaths)
+        {
+            if (importedAssets.Length == 0) return;
+
+            if (NugetHelper.NugetConfigFile == null)
+            {
+                NugetHelper.LoadNugetConfigFile();
+            }
+
+            if (NugetHelper.SettingsFile == null)
+            {
+                NugetHelper.LoadSettingFile();
+            }
+
+            var pathPrefix = NugetHelper.NugetConfigFile.RepositoryPath.Substring(
+                NugetHelper.NugetConfigFile.RepositoryPath.IndexOf(AssetsFolderName, StringComparison.Ordinal));
+            var nugetForUnityAssets = importedAssets.Where(path => path.StartsWith(pathPrefix)).ToList();
+
+            if (nugetForUnityAssets.Count == 0) return;
+
+            var runtimeAssetsFiles = nugetForUnityAssets
+                .Where(path => path.Contains(RuntimesFolderName))
+                .Where(path => path.Contains(NativeFolderName))
+                .Where(File.Exists)
+                .ToList();
+            var runtimeResults = runtimeAssetsFiles.Select(path => (path, HandleRuntime(path))).ToList();
+
+            LogResults("Handle Runtimes", runtimeResults);
+        }
+
+        private static void LogResults(string type, IList<(string, ProcessState)> results)
+        {
+            var successes = results.Where(r => ProcessState.Success.Equals(r.Item2)).Select(r => r.Item1).ToList();
+            var failures = results.Where(r => ProcessState.Failure.Equals(r.Item2)).Select(r => r.Item1).ToList();
+
+            if (successes.Count > 0 && NugetHelper.NugetConfigFile.Verbose)
+            {
+                NugetHelper.LogVerbose("NuGetForUnity: {0} successfully configured: {1}", type,
+                    string.Join(",", successes));
+            }
+
+            if (failures.Count > 0)
+            {
+                Debug.LogError($"NuGetForUnity: {type} failed to configure {string.Join(",", failures)}");
+            }
+        }
+
+        /// <summary>
+        ///     Given the <paramref name="assetFilePath" /> load the <see cref="PluginImporter" /> for the asset it's associated
+        ///     with
+        /// </summary>
+        /// <param name="assetFilePath">Path to the assets</param>
+        /// <param name="plugin">The loaded PluginImporter</param>
+        /// <returns>True if the plugin importer for the asset could be loaded</returns>
+        private static bool GetPluginImporter(string assetFilePath, out PluginImporter plugin)
+        {
+            var assetPath = assetFilePath;
+            plugin = AssetImporter.GetAtPath(assetPath) as PluginImporter;
+
+            if (plugin == null)
+            {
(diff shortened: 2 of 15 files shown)


robin-moss:
Updating on testing this:

* Linux - works
* Mac OS - Works
* Windows - Fails :( 

Some more work to set the CPU on the editor DLL, but even with this Unity complains (once) about two identical DLLs, after that one complaint it then works :thinking: so we need to figure out how to stop Unity for trying to import the two DLLs into the editor until we're ready, although that might not be possible as we are directly importing the assets.

Twometer:
In a project I'm working on it would be really useful to have this working, so I tried finding the issue:

I've noticed that for me it repeatedly doesn't work with your version, even with setting the compatibility, because the Editor settings still are at `Any CPU`. I've added the following to the importer (Windows Only), and it seems to work:
```csharp
                    if (compatibleTargets.Contains(BuildTarget.StandaloneWindows))
                    {
                        plugin.SetEditorData("CPU", "x86");
                    }
                    else if (compatibleTargets.Contains(BuildTarget.StandaloneWindows64))
                    {
                        plugin.SetEditorData("CPU", "x86_64");
                    }
                    
```
as well as doing a
```csharp
AssetDatabase.Refresh();
```
It still writes one or two "found the library twice" errors to the console, but if I ignore them and run a simple SQLite playtest:
```csharp
            var version = Marshal.PtrToStringAnsi(sqlite3_libversion());
            Assert.That(version, Is.EqualTo(_expectedVersion));
```

everything seems to work.

Maybe this helps?

robin-moss:
> In a project I'm working on it would be really useful to have this working, so I tried finding the issue:
> 
> I've noticed that for me it repeatedly doesn't work ("Compiler errors must be fixed before you can enter play mode") with your version, even with setting the compatibility, because the Editor settings still are at `Any CPU`. I've added the following to the importer (Windows Only):
> 
> ```cs
>                     if (compatibleTargets.Contains(BuildTarget.StandaloneWindows))
>                     {
>                         plugin.SetEditorData("CPU", "x86");
>                     }
>                     else if (compatibleTargets.Contains(BuildTarget.StandaloneWindows64))
>                     {
>                         plugin.SetEditorData("CPU", "x86_64");
>                     }
>                     
> ```
> 
> as well as doing a
> 
> ```cs
> AssetDatabase.Refresh();
> ```
> 
> It still writes one or two "found the library twice" errors to the console, but if I ignore them and just run a simple SQLite playtest:
> 
> ```cs
>             var version = Marshal.PtrToStringAnsi(sqlite3_libversion());
>             Assert.That(version, Is.EqualTo(_expectedVersion));
> ```
> 
> everything seems to work.
> 
> My full code is [here](https://github.com/Twometer/NuGetForUnity/tree/runtimes-support), but it's kind of messy lol
> 
> Maybe this helps?

Thanks @Twometer I will give that a shot, the Windows side is taking longer because I don't actually have access to a Windows machine most of the time.

With the current state we saw that using NuiGet to install causes the error, but it then ran fine so we've been trying to fix the test.

:facepalm: I just checked I have a local change that does exactly what you've suggested, but because we hadn't got the tests working I hadn't pushed it (will push in a minute).

robin-moss:
Tests are now working on Windows, we needed to ignore the error message to allow the tests to run.

@jwittner what are the next steps to getting this PR merged?

Twometer:
@robin-moss Could you try if it works with EmguCV?
I've tried it, and got a `DllNotFoundException` for the `cvextern.dll` when I tried it, but I don't know if the problem is with my broken branch, or if there is another issue with the importer.

robin-moss:
> @robin-moss Could you try if it works with EmguCV? I've tried that (but on Unity 2020.3), and got a `DllNotFoundException` for the `cvextern.dll` when using functions from Emgu, but I don't know if the problem is with my broken branch, or if there is another issue with the importer.

I can't easily test as I don't have a windows machine to test with, however, note that `EmguCV` also requires a runtime package to be installed as well please make sure you've done that. If you are still having a problem using it I suggest trying to call the DLL directly from C# to see if the new C# code can find the DLL, if it can then there is some magic in the package looking for a DLL with a different name (it's what gRPC does)

Twometer:
I had installed the Runtime Packages for `EmguCV`, but looking at the [source code](https://github.com/emgucv/emgucv/blob/29e8016a85a9114b5c5f6e9c3ba11247c57a52c8/Emgu.CV/PInvoke/CvInvoke.cs#L71) it seems as if they do a lot of library loading magic...

jwittner:
@robin-moss sorry I've been slammed the last couple of weeks. Next week I'll review the code, pull down the branch, do some tests and go from there. Thanks again for what's looking like an awesome contribution for a long awaited capability!

robin-moss:
Sorry to chase @jwittner, we're using a locally compiled version for now but would like to switch to the official release

jwittner:
Sorry for the delays, work and holidays took my attention away. Pulling your branch in now and I'll do some testing. =)

Assets/NuGet.config
```diff
@@ -12,4 +12,24 @@
       <add key="repositoryPath" value="./Packages" />
       <add key="DefaultPushSource" value="http://www.nuget.org/api/v2/" />
    </config>
+   <nativeRuntimesMapping>
```
jwittner, line 15: While I don't mind having these be data driven I don't think the nuget.config is the right place for them. Can you move them to a scriptable object, separate file, or simply hard code the mapping?

jwittner (changes requested):

Assets/PlayTests/NuGetPlayTests.cs
```diff
@@ -0,0 +1,76 @@
...
+            Assert.Fail("e_sqlite3 dll loaded, but should not be available");
+        }
+        catch (DllNotFoundException)
+        {
+        }
+
+        // For windows we end up importing two identical DLLs temporarily, this causes an error log that NUnit detects
+        // and would fail the test if we don't tell it to ignore the Failing messages
+        LogAssert.ignoreFailingMessages = true;
+        NugetHelper.InstallIdentifier(sqlite);
```
jwittner, line 63: I hit an exception here when running this from Unity because somehow the NugetConfigFile is not loaded, was able to fix by running `NuGetHelper.LoadNuGetConfigFile` before this call to `InstallIdentifier`.

```
NullReferenceException: Object reference not set to an instance of an object
NugetForUnity.NugetHelper.GetCachedPackage (NugetForUnity.NugetPackageIdentifier packageId) (at Assets/NuGet/Editor/NugetHelper.cs:1166)
NugetForUnity.NugetHelper.GetSpecificPackage (NugetForUnity.NugetPackageIdentifier packageId) (at Assets/NuGet/Editor/NugetHelper.cs:1112)
NugetForUnity.NugetHelper.InstallIdentifier (NugetForUnity.NugetPackageIdentifier package, System.Boolean refreshAssets) (at Assets/NuGet/Editor/NugetHelper.cs:1255)
NuGetPlayTests.Setup () (at Assets/PlayTests/NuGetPlayTests.cs:63)
UnityEditor.TestTools.TestRunner.TestLauncherBase+<>c.<ExecutePreBuildSetupMethods>b__3_0 (UnityEngine.TestTools.IPrebuildSetup targetClass) (at Library/PackageCache/com.unity.test-framework@1.1.29/UnityEditor.TestRunner/TestLaunchers/TestLauncherBase.cs:27)
UnityEditor.TestTools.TestRunner.TestLauncherBase.ExecuteMethods[T] (NUnit.Framework.Interfaces.ITest tests, NUnit.Framework.Interfaces.ITestFilter testRunnerFilter, UnityEditor.TestTools.TestRunner.AttributeFinderBase attributeFinder, System.String logString, System.Action`1[T] action, System.Nullable`1[T] testTargetPlatform) (at Library/PackageCache/com.unity.test-framework@1.1.29/UnityEditor.TestRunner/TestLaunchers/TestLauncherBase.cs:66)
UnityEditor.EditorApplication:Internal_CallUpdateFunctions()

```

Assets/PlayTests/NuGetPlayTests.cs
```diff
@@ -0,0 +1,76 @@
...
+        Assert.IsTrue(NugetHelper.IsInstalled(sqlite), "The package was NOT installed: {0} {1}", sqlite.Id,
+            sqlite.Version);
+    }
+
+    public void Cleanup()
+    {
+        // Need to reinitialise NugetHelper, but not sure why
+        NugetHelper.LoadNugetConfigFile();
+        NugetHelper.Uninstall(sqlite);
```
jwittner, line 72: I hit an exception here (if I fix the nuget config file exception) because it seems like the editor is still running. I think both of these issues may go away if you make this an edit mode test instead of a play mode one.

```
UnauthorizedAccessException: Access to the path "D:\Users\jowitt\source\External\NuGetForUnity\Assets\Packages\SQLitePCLRaw.lib.e_sqlite3.2.0.7\runtimes\win-x64\native\e_sqlite3.dll" is denied.
System.IO.File.Delete (System.String path) (at <eae584ce26bc40229c1b1aa476bfa589>:0)
System.IO.Directory.RecursiveDelete (System.String path) (at <eae584ce26bc40229c1b1aa476bfa589>:0)
System.IO.Directory.Delete (System.String path, System.Boolean recursive) (at <eae584ce26bc40229c1b1aa476bfa589>:0)
System.IO.DirectoryInfo.Delete (System.Boolean recursive) (at <eae584ce26bc40229c1b1aa476bfa589>:0)
(wrapper remoting-invoke-with-check) System.IO.DirectoryInfo.Delete(bool)
NugetForUnity.NugetHelper.DeleteDirectory (System.String directoryPath) (at Assets/NuGet/Editor/NugetHelper.cs:876)
NugetForUnity.NugetHelper.DeleteDirectory (System.String directoryPath) (at Assets/NuGet/Editor/NugetHelper.cs:862)
NugetForUnity.NugetHelper.DeleteDirectory (System.String directoryPath) (at Assets/NuGet/Editor/NugetHelper.cs:862)
NugetForUnity.NugetHelper.DeleteDirectory (System.String directoryPath) (at Assets/NuGet/Editor/NugetHelper.cs:862)
NugetForUnity.NugetHelper.Uninstall (NugetForUnity.NugetPackageIdentifier package, System.Boolean refreshAssets) (at Assets/NuGet/Editor/NugetHelper.cs:931)
NuGetPlayTests.Cleanup () (at Assets/PlayTests/NuGetPlayTests.cs:73)
UnityEditor.TestTools.TestRunner.TestLauncherBase+<>c.<ExecutePostBuildCleanupMethods>b__5_0 (UnityEngine.TestTools.IPostBuildCleanup targetClass) (at Library/PackageCache/com.unity.test-framework@1.1.29/UnityEditor.TestRunner/TestLaunchers/TestLauncherBase.cs:42)
UnityEditor.TestTools.TestRunner.TestLauncherBase.ExecuteMethods[T] (NUnit.Framework.Interfaces.ITest tests, NUnit.Framework.Interfaces.ITestFilter testRunnerFilter, UnityEditor.TestTools.TestRunner.AttributeFinderBase attributeFinder, System.String logString, System.Action`1[T] action, System.Nullable`1[T] testTargetPlatform) (at Library/PackageCache/com.unity.test-framework@1.1.29/UnityEditor.TestRunner/TestLaunchers/TestLauncherBase.cs:66)
UnityEngine.SetupCoroutine:InvokeMoveNext(IEnumerator, IntPtr)

```

jwittner (changes requested):

Assets/PlayTests/NuGetPlayTests.cs
robin-moss: Ah this is because you cannot delete a file in use on Windows, and Unity will not unload a native DLL once loaded :( 

We have to use play mode tests, we did try to use edit mode tests but we needed to force Unity to complete an asset database reload (or something like this) for it to pick up on the native files

Assets/PlayTests/NuGetPlayTests.cs
robin-moss: Thanks, will add that in. I guess because we're running with Unity open its loaded the Nuget plugin and created the file already for us

Assets/NuGet.config
robin-moss: Yeah realised that the other day, I plan to put it into a file in the `NuGet` folder and use something like Yaml/Json to make parsing easier

robin-moss: Reworked to use "settings.json" inside the "NuGet" folder

Assets/PlayTests/NuGetPlayTests.cs
robin-moss: Added the "LoadNuGetConfigFile" and the new "LoadSettingsFile" calls into the setup phase

Assets/PlayTests/NuGetPlayTests.cs
robin-moss: So this is caused by Windows not allowing you to delete a file that's in use, not sure why we didn't see it when testing on windows but we did see the need to restart Unity between tests.

I've added a catch around the deletes, that on a Windows machine will raise an error log message stating that native files cannot be deleted and a restart of Unity is needed. I'm not sure of any other solution to this I'm afraid

robin-moss:
We've been using this change locally for about three months and it's been working great, there is the known "issue" with Windows where you cannot unload C++ binaries from Unity once loaded (and cannot delete them) but I don't think it's possible to resolve this issue as it's literally a limitation of Windows

Twometer:
It is possible to unload C++ binaries from a Windows application with the `FreeLibrary` call, but I don't know if/how that works in the context of Unity.

robin-moss:
To use that we would need to make setup OS-specific native DllImports so we could unload the native libraries (assuming this doesn't break Unity Editor).

This is only really an issue because Windows doesn't let us delete files that are in use :(

robrab2000:
Hello :) I was curious if there was likely to be any movement on this getting merged?

JoC0de:
@robrab2000 It is the biggest think left "on the list" but it is not trivial. We can add the support as it is implemented in this branch. But it will never work for "every" package. Packages that have native binaries often also have custom import steps (MSBuild Tasks) that we cant run. So we can't ensure that the NuGet package will work inside Unity. E.g. the nuget package need to contain one native binary for each target platform if one is missing it wouldn't work on the device. So it is generally not a good think to import packages with nativ dependencies. I am not sure how we can make this clear so users don't get into the "It works on my Machine" issue.

robrab2000:
If there is an implementation which works for some packages and not for others, could it not be like an 'experimental' flag in the settings? Regarding platforms, perhaps you could just indicate beforehand what platform you would like. in fact perhaps there could be a separate runtime extraction tool where you can extract a runtime for your platform (one at a time so that there is no conflict between them before you get a chance to select a platform for it in the inspector).

Maybe I'm being silly and there are good reasons that these approaches wouldn't work?

LupusUmbrae:
We've been using a fork of this repo with the native stuff in for quite awhile.. mostly for gRPC support though.

A better way to do this is to use an asset import processor that can prevent the import before unity even tries to load it. As well as doing things like disabling auto importing be default.

But that's also a major departure from how this tool works (or worked).

My eventual plan was to replace the tool completely with just an asset processor and real nuget cli but just haven't had the time.

JoC0de:
@robrab2000 Yes more or less we can implement it and show a warning so the User is aware of the fact that it contains native dependencies, and listing all platforms that are included inside the NuGet package so he is aware of possible missing support of target platforms.

chubei-urus:
Hi @robin-moss may I ask what's the fork that you've been using? I checked https://github.com/robin-moss/NuGetForUnity but it's quite some commits behind so I'm not sure if it's the correct one.

robin-moss:
That is the fork we're using, I hadn't updated against this repo in quite some time and looks like it's gone through some serious changes making it much harder to rebase my changes :( 

I don't think I'll rebase this change on the latest version even if I want V3 Nuget support, just a lot of effort for something that is currently working ok for us

Issue #362, opened 2021-01-13
Azure Artifacts Custom Package Error


zkkzkk32312:
## Description

I am trying to add custom Azure Artifacts Nuget package without success.
I have edited/added the \<packageSources> and \<packageSourceCredentials> in the config file.
but I am getting "Unable to retrieve package list...." Error in the Unity console when I hit the refresh button.
Any suggestion would be appreciated. 

...
\<packageSources>
     ...
     \<add key="ABC" value="https://pkgs.dev.azure.com/....json" />
     ...
\</packageSources>
...
\<packageSourceCredentials>
	\<ABC>
		\<add key="Username" value="ABC@ABC.com" />
		\<add key="Password" value="1234567" />
	\</ABC>
\</packageSourceCredentials>
...

System.Net.WebException: The remote server returned an error: (404) Not Found.
  at System.Net.HttpWebRequest.EndGetResponse (System.IAsyncResult asyncResult) [0x00058] in <ef151b6abb5d474cb2c1cb8906a8b5a4>:0 
  at System.Net.HttpWebRequest.GetResponse () [0x0000e] in <ef151b6abb5d474cb2c1cb8906a8b5a4>:0 
  at NugetForUnity.NugetHelper.RequestUrl (System.String url, System.String userName, System.String password, System.Nullable`1[T] timeOut) [0x000b9] in <b7bde984cef1447da61a3fd28d4789b0>:0 
  at NugetForUnity.NugetPackageSource.GetPackagesFromUrl (System.String url, System.String username, System.String password) [0x0006c] in <b7bde984cef1447da61a3fd28d4789b0>:0 
  at NugetForUnity.NugetPackageSource.Search (System.String searchTerm, System.Boolean includeAllVersions, System.Boolean includePrerelease, System.Int32 numberToGet, System.Int32 numberToSkip) [0x000ce] in <b7bde984cef1447da61a3fd28d4789b0>:0 
UnityEngine.Debug:LogErrorFormat(String, Object[])
NugetForUnity.NugetPackageSource:Search(String, Boolean, Boolean, Int32, Int32)
NugetForUnity.NugetHelper:Search(String, Boolean, Boolean, Int32, Int32)
NugetForUnity.NugetWindow:UpdateOnlinePackages()

- **NuGet Package:** [Custom Azure Artifacts Nuget Package]
- **NuGetForUnity Version:** [2.0.1]
- **Unity Version:** [2019.4]
[...]

PR #422, opened 2021-11-01
Support hyphen in Azure org account name

ccarvalheiro:
As recently discussed on #362, we would fail to get the correct provider when an Azure feed pointed to an organization with a hyphen in its name.

When creating an organization under Azure DevOps, the following rules for naming apply:
- Should start with a letter or number only;
- Should end with a letter or number only;
- In between, it can only contain letters, numbers, and hyphens.

This commit adds the hyphen character to the list of supported ones in the regex that is used to look for the account name in the feed URL. Since the rules are enforced on creation, I didn't add any complexity to the regex by trying to follow the naming convention.

I don't know if the same applies to the other rule, but in case anyone does I can add easily patch it too.

### DIFF
--- a/Assets/NuGet/Editor/NugetHelper.cs
+++ b/Assets/NuGet/Editor/NugetHelper.cs
@@ -1492,7 +1492,7 @@
         {
             new AuthenticatedFeed()
             {
-                AccountUrlPattern = @"^https:\/\/(?<account>[a-zA-z0-9]+).pkgs.visualstudio.com",
+                AccountUrlPattern = @"^https:\/\/(?<account>[-a-zA-z0-9]+).pkgs.visualstudio.com",
                 ProviderUrlTemplate = "https://{account}.pkgs.visualstudio.com/_apis/public/nuget/client/CredentialProviderBundle.zip"
             },
             new AuthenticatedFeed()
--- a/Assets/NuGet/Editor/NugetHelper.cs
+++ b/Assets/NuGet/Editor/NugetHelper.cs
@@ -1497,7 +1497,7 @@
             },
             new AuthenticatedFeed()
             {
-                AccountUrlPattern = @"^https:\/\/pkgs.dev.azure.com\/(?<account>[a-zA-z0-9]+)\/",
+                AccountUrlPattern = @"^https:\/\/pkgs.dev.azure.com\/(?<account>[-a-zA-z0-9]+)\/",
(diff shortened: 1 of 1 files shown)


Assets/NuGet/Editor/NugetHelper.cs
```diff
@@ -1497,7 +1497,7 @@ public string GetProviderUrl(string account)
             },
             new AuthenticatedFeed()
             {
-                AccountUrlPattern = @"^https:\/\/pkgs.dev.azure.com\/(?<account>[a-zA-z0-9]+)\/",
+                AccountUrlPattern = @"^https:\/\/pkgs.dev.azure.com\/(?<account>[-a-zA-z0-9]+)\/",
```
jwittner, line 1500: There's another AuthenticatedFeed in this list with a similar 'account' defined in it's regex. Can you fix that one too?

jwittner (changes requested):
Looking great, just would like the fix for all the url patterns we know of. =)

Assets/NuGet/Editor/NugetHelper.cs
ccarvalheiro: Sure, updated.

ccarvalheiro: Do you plan on publishing a new release with this fix?

jwittner:
Hey @ccarvalheiro - I've made a [PR](https://github.com/ccarvalheiro/NuGetForUnity/pull/1) into your branch that adds a couple more fixes to the regex I found testing your branch. If you can review/approve/merge that PR into your branch I'm a go on merging this into main.

I can make a new release once these changes are in. =)

Assets/NuGet/Editor/NugetHelper.cs
jwittner: I can make a new release once these changes are in. =)

ccarvalheiro:
Hey @jwittner, just merged. Sorry for missing those improvements. Thanks

jwittner (approved):

jwittner:
@ccarvalheiro - released as [v3.0.3](https://github.com/GlitchEnzo/NuGetForUnity/releases/tag/v3.0.3).  Thanks for the contributions!

ccarvalheiro:
Cool! Thank you!

Issue #407, opened 2021-07-05
Any way to disable automatic restore on each recompile?


lehtiniemi:
## Description
Is there any way to disable restoring packages on every recompile (and maybe just do it on Unity start)? With just a couple NuGet-packages installed, Unity restores them on each recompile and this is huge amount of recompile time per day.

- **NuGet Package:** [What is the name and version of the NuGet package you are installing?]
![image](https://user-images.githubusercontent.com/13351978/124436077-4e0c9f80-dd7e-11eb-9487-e9a0d53489ad.png)

- **NuGetForUnity Version:** [What version of NuGetForUnity are you running?]
- 3.0.2
- **Unity Version:** [What version of Unity are you using?]
- 2021.2.0a16
- **Operating System:** [What OS are you on?]
- OSX 10.15.7

jwittner:
Hmmm, it should be detecting that they're already installed and skipping any installation and avoiding any recompiling by doing so. I wonder if this is only on OSX?

NoTuxNoBux:
> I wonder if this is only on OSX?

Negative, I can confirm I'm also experiencing it on Windows with Unity 2020.3.11f1 :slightly_smiling_face: .

I don't recall it happening with Unity 2019.4.27, but Unity 2019 didn't show a progress bar with what it was doing on recompile, so it may well have been.

AlexanderLuck:
We just got the package a week ago and have the same issue on Windows. 
Just 2 packages adds 4-5 seconds every time we compile, which makes using this package hard to justify.
Is this being worked on?

jwittner:
That does seem excessive for determining that the packages required are already installed. No one is looking into this at the moment that I know of.

Would love:
* an investigation into why it's slow.
* a personal preference in Editor -> Preferences to control the automatic restore.

PRs are welcome! =)

Shii2:
As a workaround for now I added variable:
`private const string k_ProjectOpened = "ProjectOpened";` to the NugetHelper.cs
and guarded Restore with this code:
```
if (!SessionState.GetBool(k_ProjectOpened, false))
{
    SessionState.SetBool(k_ProjectOpened, true);
    Restore();
}
```
After this change Nuget only calls Restore one time after loading the project and not calling after changing any scripts in the project.

NoTuxNoBux:
[...]

PR #424, opened 2021-11-10
Fix package restore happening on every script recompilation in projects

NoTuxNoBux:
When NuGetForUnity is included in a user project, and they make changes to their scripts, Unity also reloads the editor scripts for NuGetForUnity, which will then retrigger the static NugetHelper constructor because it is marked as InitializeOnLoad. Because this constructor restores packages, a package restore is done every script reload. Usually this amounts to no actual change because the packages in question are already installed, but just checking one or two packages this way can already slow down the script recompilation process by several seconds, which quickly becomes tiresome.

This changes the restore to only happen once on project start-up, fixing the issue. There was also little benefit to the restore happening, as it didn't happen automatically when packages.config was modified, which would have been the most useful and already requires a manual reload.

Solution was proposed by @Shii2 in #407. I just formalized the change in a PR, also skipped doing I/O to check if the packages folder exists every time, and tested in a Unity 2020.3.20f1 project that it works as intended. Checks done:

- [x] Starting up a project still restores packages.
- [x] Making a change to a script no longer restores packages, even multiple times in succession.

### DIFF
--- a/Assets/NuGet/Editor/NugetHelper.cs
+++ b/Assets/NuGet/Editor/NugetHelper.cs
@@ -110,12 +110,12 @@
             insideInitializeOnLoad = true;
             try
             {
-                if (SessionState.GetBool("FirstProjectOpen", false))
+                if (SessionState.GetBool("NugetForUnity.FirstProjectOpen", false))
                 {
                     return;
                 }
 
-                SessionState.SetBool("FirstProjectOpen", true);
+                SessionState.SetBool("NugetForUnity.FirstProjectOpen", true);
                 
                 // if we are entering playmode, don't do anything
                 if (EditorApplication.isPlayingOrWillChangePlaymode)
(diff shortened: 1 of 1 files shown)


Assets/NuGet/Editor/NugetHelper.cs
```diff
@@ -110,6 +110,13 @@ static NugetHelper()
             insideInitializeOnLoad = true;
             try
             {
+                if (SessionState.GetBool("FirstProjectOpen", false))
```
jwittner, line 113: Since the `SessionState` is global can we modify the key to help avoid any clashes? Maybe `NugetForUnity.FirstProjectOpen`?

jwittner (changes requested):
One small change request, otherwise this looks like a decent solution.

Assets/NuGet/Editor/NugetHelper.cs
NoTuxNoBux: Agreed, and done!

jwittner (approved):

PR #429, opened 2021-11-29
fix: encode packageSourceCredentials element names

robin-moss:


jwittner:
Works great! Thanks a ton for the tests!

Issue #395, opened 2021-05-10
Unable to install nuget packages that exists as part of MonoBleedingEdge


jmfGE:
## Description
I'm trying to install RabbitMQ.Client 6.2.1, but NuGetForUnity debug says the library is already installed. I drilled down and found it was finding that RabbitMQ is part of the library for the Unity Editor (under MonoBleedingEdge), but those libraries are not actually exposed to the Unity runtime environment. 

"C:\Program Files\Unity\Hub\Editor\2021.1.2f1\Editor\Data\MonoBleedingEdge\lib\mono\4.7.1-api\RabbitMQ.Client.dll"

- **NuGet Package:** RabbitMQ.Client 6.2.1
- **NuGetForUnity Version:** 3.0.2
- **Unity Version:** 2021..2f1
- **Operating System:** Windows 10

jwittner:
Was worried this day would come. We need to either find out a way to determine dynamically what libraries in those folders are/aren't exposed to the runtime or we need to make a whitelist of ones in those folders to ignore in the test. The list of assemblies we're assuming are already included is gather in [this code](https://github.com/GlitchEnzo/NuGetForUnity/blob/005704459f3327230dbd570d405f0bb5437e4a0c/Assets/NuGet/Editor/NugetHelper.cs#L555).

Open to ideas for building up that list more intelligently / blacklisting assemblies from inclusion in that list even if they are under the folders. If we know there are whole folders we can ignore that'd be a great start!

robin-moss:
Could we not just exclude the `MonoBleedingEdge` folder from the search?

or could we use `AppDomain.CurrentDomain.GetAssemblies()` to find all active domains in Unity?

robin-moss:
Looks like `AppDomain.CurrentDomain.GetAssemblies().Select(a => a.ManifestModule.Name)
                    .Select(p => Path.ChangeExtension(p, null));` should work :)

PR #431, opened 2021-12-21
fix: use reflection to find already imported libs #395

robin-moss:


Assets/NuGet/Editor/NugetHelper.cs
```diff
@@ -554,12 +554,11 @@ private static bool IsAlreadyImportedInEngine(NugetPackageIdentifier package)
...
         private static HashSet<string> GetAlreadyImportedLibs()
         {
+            
             if (alreadyImportedLibs == null)
             {
-                string[] lookupPaths = GetAllLookupPaths();
-                IEnumerable<string> libNames = lookupPaths
-                    .SelectMany(directory => Directory.EnumerateFiles(directory, "*.dll", SearchOption.AllDirectories))
-                    .Select(Path.GetFileName)
+                IEnumerable<string> libNames = AppDomain.CurrentDomain.GetAssemblies()
```
jwittner, line 560: A little worried that this will catch assemblies we've installed with NuGet. For example, if you install a package, then try to update it, will this find the previous version and fail to update the assembly?

JoC0de:
I rebased and reformatted your code. It works as far as I can see (I could also remove a workaround needed for installing System.Text.Json)👍
So thank you @robin-moss for the improvement.

PR #469, opened 2022-10-16
Change Project Hierarchy to import it as Unity custom package. 

AwesomeGameConcepts:
Your code is very useful for me but i'm facing some limitation.
Because the code is in project organization and can't be imported easily in Unity (I have my own Project structure,  i can't remove / add / update your code with unity Package Manager.
So I simply move your code into a CustomPackage for a more project friendly usage (https://docs.unity3d.com/Manual/CustomPackages.html).
This way i can go to the Package manager and click on add package from git, put the git repo address (https://github.com/AwesomeGameConcepts/NuGetForUnity.git), and it will be automatically add your code and needed reference (not set but easy to add).
image

This way i can have your code in any unity project (without disturb actual project hierarchy), and manage delete/add/update behavior from Package Manager.

I also change somme deprecated function in Unity 2021.3

Thanks a lot :)


JoC0de:
Used this PR as a basis to implement it in #480 and #483.
@AwesomeGameConcepts Thanks for the contribution (inspiration)

PR #480, opened 2023-01-09
restructure repository and support UPM

JoC0de:
restructure repository
add support for UPM


JoC0de:
@jwittner can you have a short look at the new repository structure (just look at the resulting https://github.com/JoC0de/NuGetForUnity/tree/feature/update-project-layout)?
It is not required to change the repository structure to support UPM I just wanted to "clean up" the structure because wen we support UPM we can't change (at least the part where the `package.json` is located).
For easier understanding:
- `src/NuGetForUnity/` is the UPM package, it contains only the code that is needed by the "user"
- `src/NuGetForUnity.Tests/` is the Unity project that contains the tests, it includes NuGetForUnity as a local package.

jwittner:
Looks reasonable!

JoC0de:
@jwittner  Ok thanks for the feedback

PR #491, opened 2023-01-26
Skip .svn subdirectory

doctaweeks:
Older SVN versions will create a .svn subdirectory in each directory of the repo. If one is present in the RepositoryPath, skip it, otherwise it will be removed, damaging the repo.

### DIFF
--- a/src/NuGetForUnity/Editor/NugetHelper.cs
+++ b/src/NuGetForUnity/Editor/NugetHelper.cs
@@ -1403,12 +1403,14 @@ namespace NugetForUnity
             var directories = Directory.GetDirectories(NugetConfigFile.RepositoryPath, "*", SearchOption.TopDirectoryOnly);
             foreach (var folder in directories)
             {
-                if (Path.GetFileName(folder) == ".svn")
+                var folderName = Path.GetFileName(folder);
+                if (folderName.Equals(".svn", StringComparison.OrdinalIgnoreCase))
                 {
+                    // ignore folder required by SVN tool
                     continue;
                 }
 
-                var pkgPath = Path.Combine(folder, $"{Path.GetFileName(folder)}.nupkg");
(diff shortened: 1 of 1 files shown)


JoC0de:
The `RepositoryPath` is the folder where all nuget packages are installed, so it is a folder that is created and managed by a Package Manager. So it shouldn't be pushed to a version control system like GIT or SVN. And when you don't push it you should not have the problem. Or am I wrong?

doctaweeks:
You are correct, if it not checked in, then it won't be a problem. However, there are reasons to check packages in, for example:

* Builds where there is no network access
* Reduce external dependencies in a build pipeline
* Cut down first build time on a fresh checkout

src/NuGetForUnity/Editor/NugetHelper.cs
```diff
@@ -1403,6 +1403,11 @@ internal static void CheckForUnnecessaryPackages()
             var directories = Directory.GetDirectories(NugetConfigFile.RepositoryPath, "*", SearchOption.TopDirectoryOnly);
             foreach (var folder in directories)
             {
+                if (Path.GetFileName(folder) == ".svn")
+                {
+                    continue;
+                }
+
                 var pkgPath = Path.Combine(folder, $"{Path.GetFileName(folder)}.nupkg");
```
JoC0de, line 1411:
```suggestion
                var folderName = Path.GetFileName(folder);
                if (folderName.Equals(".svn", StringComparison.OrdinalIgnoreCase))
                {
                    continue;
                }

                var pkgPath = Path.Combine(folder, $"{folderName}.nupkg");
```

JoC0de (changes requested):

JoC0de:
Ok you are right. If you 'apply' my suggestion I will merge it.

PR #510, opened 2023-04-07
New feature - selection

popara96:
Adding the ability to select packages that the user wants to uninstall or update (instead of uninstalling/updating all). Also added an option to uninstall all packages.


src/NuGetForUnity/Editor/NugetHelper.cs
```diff
@@ -72,6 +72,11 @@ public static class NugetHelper
         /// </summary>
         private static readonly Dictionary<string, NugetPackage> installedPackages = new Dictionary<string, NugetPackage>();
 
+        /// <summary>
+        ///     The dictionary of currently selected installed <see cref="NugetPackage" />s keyed off of their ID string.
+        /// </summary>
+        private static readonly Dictionary<string, NugetPackage> selectedInstalledPackages = new Dictionary<string, NugetPackage>();
```
igor84, line 78: I think you can keep this is List instead of Dictionary. You are only using it by key in Unselect method but you are only calling that method when you determine the item among the Values of this dict so everything could be made to work with List just the same.

src/NuGetForUnity/Editor/NugetHelper.cs
```diff
@@ -589,6 +602,39 @@ public static void Push(NuspecFile nuspec, string nuspecFilePath, string apiKey
...
         }
 
+        /// <summary>
+        /// Selects the given package to be uninstalled or updated.
+        /// </summary>
+        /// <param name="package">Package to be selected.</param>
+        public static void Select(NugetPackage package)
+        {
+            selectedInstalledPackages.Add(package.Id, package);
+            SelectedPackagesCount++;
```
igor84, line 612: It would be clearer if you would set package.Selected to true in here and set it to false in Unselect method.

src/NuGetForUnity/Editor/NugetHelper.cs
```diff
@@ -709,6 +755,19 @@ internal static void UninstallAll()
...
             }
+            UnselectAll();
+        }
+
+        /// <summary>
+        ///     Uninstalls all of the currently installed and selected packages.
+        /// </summary>
+        internal static void UninstallAllSelected()
+        {
+            foreach (var package in selectedInstalledPackages.Values.ToList())
```
igor84, line 766: You should be able to write this as a for loop that just calls Uninstall for each element and then after the loop you could just call UnselectAll().

src/NuGetForUnity/Editor/NugetWindow.cs
```diff
@@ -724,11 +746,24 @@ private void DrawUpdatesHeader()
...
+                            NugetHelper.UpdateInstalledPackages();
+                            UpdateUpdatePackages();
+                        }
+                    }
+
+                    if (NugetHelper.SelectedPackagesCount > 0)
+                    {
+                        if (GUILayout.Button("Install All Selected Updates", GUILayout.Width(200)))
+                        {
+                            NugetHelper.UpdateAllSelected(updatePackages);
```
igor84, line 763: It looks like you didn't have to add new UpdateAllSelected method. You could have just called this:
```c#
NugetHelper.UpdateAll(selectedInstalledPackages, NugetHelper.InstalledPackages);
NugetHelper.UnselectAll();
```

igor84 (changes requested):
We would also want to be able to select packages that we want to install but it is good that this commit focuses on only on updates and uninstalls for now since install would probably require a bit better UX.

src/NuGetForUnity/Editor/NugetHelper.cs
```diff
@@ -137,6 +142,14 @@ public static PackagesConfigFile PackagesConfigFile
         /// <returns>A dictionary of installed <see cref="NugetPackage" />s.</returns>
         public static IEnumerable<NugetPackage> InstalledPackages => installedPackages.Values;
 
+        /// <summary>
+        ///     Gets the dictionary of packages that are actually installed in the project and selected, keyed off of the ID.
+        /// </summary>
+        /// <returns>A dictionary of installed and selected <see cref="NugetPackage" />s.</returns>
+        public static IEnumerable<NugetPackage> SelectedInstalledPackages => selectedInstalledPackages.Values;
+
+        public static int SelectedPackagesCount;
```
JoC0de, line 151: Why not just do `selectedInstalledPackages.Count`?

src/NuGetForUnity/Editor/NugetHelper.cs
JoC0de, line 78: I know the code of `NuGetForUnity` is not the most clean but pleas try not to pollute the `NugetHelper` class. As far as I see you only need this inside the UI (`NugetWindow`) so I don't see a reason to place this inside the "global" `NugetHelper` class. Also you are "duplicating" the information you have a collection of selected + a bool on each package. Duplicating is normally not a good practice as it can get out of sync. I think we should only use one collection and no bool as you are storing a state of a single UI so it should not interfere with other code. I think you already have the issue that you don't know if the user selected a package for uninstall or for updating as you store both in the same location.

JoC0de (changes requested):

src/NuGetForUnity/Editor/NugetWindow.cs
```diff
@@ -683,39 +678,29 @@ private void DrawInstalledHeader()
                         }
                     }
 
-                    if (NugetHelper.SelectedPackagesCount > 0)
+                    if (NugetHelper.InstalledPackages.Where(p => p.IsSelected).Count() > 0)
```
igor84, line 681: I think Count can accept a predicate as well so this could be `InstalledPackages.Count(p => p.IsSelected)`

src/NuGetForUnity/Editor/NugetWindow.cs
```diff
@@ -62,7 +62,22 @@ public class NugetWindow : EditorWindow
         /// <summary>
         ///     The filtered list of package updates available.
         /// </summary>
-        private List<NugetPackage> filteredUpdatePackages = new List<NugetPackage>();
+
+        private List<NugetPackage> FilteredUpdatePackages
+        {
+            get
+            {
+                if (updatesSearchTerm == "Search")
```
igor84, line 70: Add `|| string.IsNullOrWhitepsace(updatesSearchTerm)` here.

src/NuGetForUnity/Editor/NugetWindow.cs
```diff
@@ -748,34 +733,25 @@ private void DrawUpdatesHeader()
...
                             NugetHelper.UpdateAll(updatePackages, NugetHelper.InstalledPackages);
                             NugetHelper.UpdateInstalledPackages();
                             UpdateUpdatePackages();
                         }
-                    }
 
-                    if (NugetHelper.SelectedPackagesCount > 0)
-                    {
-                        if (GUILayout.Button("Install All Selected Updates", GUILayout.Width(200)))
+                        if (updatePackages.Where(p => p.IsSelected).Count() > 0)
```
igor84, line 743: Same for Count predicate.

src/NuGetForUnity/Editor/NugetWindow.cs
JoC0de:
```suggestion
                    if (NugetHelper.InstalledPackages.Any(package => package.IsSelected))
```
actually you don't even need to count the packages.

src/NuGetForUnity/Editor/NugetWindow.cs
JoC0de:
```suggestion
                        if (updatePackages.Any(package => package.IsSelected))
```

src/NuGetForUnity/Editor/NugetWindow.cs
```diff
@@ -724,23 +731,27 @@ private void DrawUpdatesHeader()
...
+                        }
 
-                    if (GUILayout.Button("Refresh", GUILayout.Width(60)))
-                    {
-                        Refresh(true);
+                        if (updatePackages.Where(p => p.IsSelected).Count() > 0)
+                        {
+                            if (GUILayout.Button("Update Selected", GUILayout.Width(200)))
+                            {
+                                NugetHelper.UpdateAll(updatePackages.Where(p => p.IsSelected), NugetHelper.InstalledPackages);
```
JoC0de, line 747: I still think a private field `private readonly HashSet<NugetPackage> selectedPackages = new HashSet<NugetPackage>();`. Then wenn you need the selected packages you need to do: `updatePackages.Where(selectedPackages.Contains)`. The reason for this is just to separate UI state from logic. E.g. when refreshing the data (updatePackages) you don't automatically lose the selection state (what package the user has selected).

igor84: Ah, right. Just to clearify: it should be better if there is no IsSelected field in NugetPackage itself since that is view only data and should thus only exist in the view layer and one way to save what are the selected packages in view layer is to use `HashSet<NugetPackage>` there. Is that right?

JoC0de: Exactly

popara96:
Small note about the most recent commit: IsSelected was removed when I merged master into select_packages branch, as I decided to resolve it that way because I was planning to remove the flag in the coming change.

src/NuGetForUnity/Editor/NugetWindow.cs
```diff
@@ -672,17 +664,17 @@ private void DrawInstalledHeader()
...
                         }
                     }
 
-                    if (NugetHelper.InstalledPackages.Where(p => p.IsSelected).Count() > 0)
+                    if (NugetHelper.InstalledPackages.Any(selectedPackages.Contains))
                     {
                         if (GUILayout.Button("Uninstall Selected", EditorStyles.miniButtonRight, GUILayout.Width(150)))
                         {
-                            NugetHelper.UninstallAll(true);
+                            NugetHelper.UninstallAll(NugetHelper.InstalledPackages.Where(p => selectedPackages.Contains(p)).ToList());
```
igor84, line 677: Can you just ask `if (selectedPackages.Count > 0)` above and just do `NugetHelper.UninstallAll(selectedPackages.ToList())` here?

src/NuGetForUnity/Editor/NugetWindow.cs
```diff
@@ -661,39 +657,42 @@ private void DrawInstalledHeader()
 
             EditorGUILayout.BeginVertical(headerStyle);
             {
-                var enterPressed = Event.current.Equals(Event.KeyboardEvent("return"));
-
                 EditorGUILayout.BeginHorizontal();
                 {
-                    if (GUILayout.Button("Preferences", GUILayout.Width(80)))
+                    GUILayout.FlexibleSpace();
+                    if (NugetHelper.InstalledPackages.Count() > 0)
```
JoC0de, line 663:
```suggestion
                    if (NugetHelper.InstalledPackages.Any())
```

src/NuGetForUnity/Editor/NugetWindow.cs
JoC0de: @igor84 I think it is safer to `NugetHelper.UninstallAll(NugetHelper.InstalledPackages.Where(selectedPackages.Contains).ToList())` because it protects against having a package that was already uninstalled in the `selectedPackages` list. The `selectedPackages` is managed by the UI but the UI doesn't know what the `NugetHelper` dose. So I would go the safe way of only trusting the state managed by `NugetHelper`.

src/NuGetForUnity/Editor/NugetWindow.cs
```diff
@@ -724,23 +723,27 @@ private void DrawUpdatesHeader()
...
+                        }
 
-                    if (GUILayout.Button("Refresh", GUILayout.Width(60)))
-                    {
-                        Refresh(true);
+                        if (updatePackages.Any(selectedPackages.Contains))
+                        {
+                            if (GUILayout.Button("Update Selected", GUILayout.Width(200)))
+                            {
+                                NugetHelper.UpdateAll(updatePackages.Where(p => selectedPackages.Contains(p)), NugetHelper.InstalledPackages);
```
JoC0de, line 739:
```suggestion
                                NugetHelper.UpdateAll(updatePackages.Where(selectedPackages.Contains), NugetHelper.InstalledPackages);
```
It's the same but just for "style" 😄 .

JoC0de:
@popara96 I now like it much more. Are there still thinks you need to work on?

popara96:
@JoC0de I'll fix the unit tests theses days when I have time, probably tomorrow. As for what I need to work on, I'd like to do similar thing for 'Online' tab but I will probably do it in a separate PR because there might be additional changes in order to do it.

src/NuGetForUnity/Editor/NugetWindow.cs
JoC0de:
```suggestion
                            NugetHelper.UninstallAll(NugetHelper.InstalledPackages.Where(selectedPackages.Contains).ToList());
```

JoC0de (approved):
I fixed the last thinks by myself.
I will merge it now, so thanks for the contribution 👍.

PR #514, opened 2023-04-16
document how to resolve missing System libraries

### LINKED ISSUE #366: PDF Sharp 'FontFamily' is defined in an assembly that is not referenced
## Description
I'm getting an error CS0012: The type 'FontFamily' is defined in an assembly that is not referenced. You must add a reference to assembly 'System.Drawing, Version=2.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'.

I can make this go away in visual studio with the "add reference" action but this only solves the issue in visual studio, it still shows up in Unity's console
image

before:
image

after:
image

- **NuGet Package:** PdfSharp Version 1.50.5147
- **NuGetForUnity Version:** 2.0.1
- **Unity Version:** 2020.2.1

JoC0de:
document how to resolve missing System libraries

Closes #366

### DIFF
--- a/README.md
+++ b/README.md
@@ -235,4 +235,4 @@ To import one of the System libraries and resolve the error you need to create a
 -r:System.Net.Http.dll
 ```
 
-and place it inside the containig the project that requires the library e.g. the `Assets` folder. It can also be placed in a folder that contains a `.asmdef` so the reference is only added to the sub-project. NuGetForUnity also uses a `csc.rsp` file see [csc.rsp](src/NuGetForUnity/Editor/csc.rsp).
+and place it inside the containing project that requires the library e.g. the `Assets` folder. It can also be placed in a folder that contains a `.asmdef` to only add the reference to the sub-project. For example NuGetForUnity also uses a `csc.rsp` file see [csc.rsp](src/NuGetForUnity/Editor/csc.rsp).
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -220,3 +220,19 @@ As mentioned in the message this error can be suppressed by changing a setting.
...
+The type `HttpClient` is defined in an assembly that is not referenced. You must add a reference to assembly 'System.Net.Http, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'.
+```
+
+To import one of the System libraries and resolve the error you need to create a `csc.rsp` file with the content e.g. to import `System.Net.Http`
+
+```
+-r:System.Net.Http.dll
+```
+
+and place it inside the containig the project that requires the library e.g. the `Assets` folder. It can also be placed in a folder that contains a `.asmdef` so the reference is only added to the sub-project. NuGetForUnity also uses a `csc.rsp` file see [csc.rsp](src/NuGetForUnity/Editor/csc.rsp).
```
JoC0de, line 238:
```suggestion
and place it inside the containing project that requires the library e.g. the `Assets` folder. It can also be placed in a folder that contains a `.asmdef` to only add the reference to the sub-project. For example NuGetForUnity also uses a `csc.rsp` file see [csc.rsp](src/NuGetForUnity/Editor/csc.rsp).
```

PR #518, opened 2023-04-25
Allow disabling installing dependencies on restore

tete1030:
NugetForUnity is constantly adding and installing a package, which was explicitly deleted from packages.config, when opening Unity Editor after initial clone. This package (`System.Threading.Tasks.Extensions`) is a dependency of another. It has duplicated definition of `ValueTask` class with netstandard. Deleting the package is thus necessary and has not caused any issue for me, so I have adopted this solution.

I think respecting instead of updating `packages.config` when restoring is a reasonable case. For this reason I added an option `Lock packages on restore` to disable installing dependencies of a restoring package.

I also added a timeout option for nuget mirror requests. We have used customized nuget proxy with longer http request than usual (almost 20s).

Result:
image


JoC0de:
Hi @tete1030 thanks for your contribution.
Can you please describe your issue with `System.Threading.Tasks.Extensions` as the option `Lock packages on restore` is only a workaround. Normally packages that are 'not required' should be handled by `UnityPreImportedLibraryResolver` so NuGetForUnity doesn't even try to install them. Do you have a example NuGet package that produces the error? What version of Unity do you use? What is your target framework (Api Compatibility Level)?

tete1030:
I agree that it's a workaround for the `System.Threading.Tasks.Extensions` issue alone. But for this option itself it provides more. For example it creates stable and reproducible environments and same env with CLI restoring process.

You can try the `fo-dicom 5.0.3` package. We are using Unity 2022.2.2f1 and the API compatibility level is .net standard 2.1. First add some script with `ValueTask` class.

The problem cannot be seen from Unity. It compiles well without any error. However when we open the project with VS2019 or VS2022 and try to attach to unity, the error appears in VS.

tete1030:
Do you need further information on this problem?

src/NuGetForUnity/Editor/NugetHelper.cs
```diff
@@ -1075,7 +1067,7 @@ private static void CopyStream(Stream input, Stream output)
         /// </summary>
         /// <param name="package">The identifier of the package to install.</param>
         /// <param name="refreshAssets">True to refresh the Unity asset database.  False to ignore the changes (temporarily).</param>
-        internal static bool InstallIdentifier(NugetPackageIdentifier package, bool refreshAssets = true)
+        internal static bool InstallIdentifier(NugetPackageIdentifier package, bool refreshAssets = true, bool installDeps = true)
```
JoC0de, line 1070:
```suggestion
        internal static bool InstallIdentifier(NugetPackageIdentifier package, bool refreshAssets = true, bool installDependencies = true)
```

src/NuGetForUnity/Editor/NugetHelper.cs
```diff
@@ -1075,7 +1067,7 @@ private static void CopyStream(Stream input, Stream output)
         /// </summary>
         /// <param name="package">The identifier of the package to install.</param>
         /// <param name="refreshAssets">True to refresh the Unity asset database.  False to ignore the changes (temporarily).</param>
```
JoC0de, line 1069:
```suggestion
        /// <param name="refreshAssets">True to refresh the Unity asset database.  False to ignore the changes (temporarily).</param>
        /// <param name="installDependencies">True to also install all dependencies of the <paramref name="package" />.</param>
```

src/NuGetForUnity/Editor/NugetHelper.cs
```diff
@@ -1115,7 +1107,7 @@ public static void LogVerbose(string format, params object[] args)
         /// </summary>
         /// <param name="package">The package to install.</param>
         /// <param name="refreshAssets">True to refresh the Unity asset database.  False to ignore the changes (temporarily).</param>
-        public static bool Install(NugetPackage package, bool refreshAssets = true)
+        public static bool Install(NugetPackage package, bool refreshAssets = true, bool installDeps = true)
```
JoC0de, line 1110:
```suggestion
        public static bool Install(NugetPackage package, bool refreshAssets = true, bool installDependencies = true)
```

src/NuGetForUnity/Editor/NugetHelper.cs
```diff
@@ -1373,7 +1365,7 @@ public static Stream RequestUrl(string url, string userName, string password, in
         /// <summary>
         ///     Restores all packages defined in packages.config.
         /// </summary>
-        public static void Restore()
+        public static void Restore(bool installDeps = true)
```
JoC0de, line 1368:
```suggestion
        public static void Restore(bool installDependencies = true)
```

src/NuGetForUnity/Editor/NugetConfigFile.cs
```diff
@@ -62,6 +62,21 @@ public class NugetConfigFile
         /// </summary>
         public bool ReadOnlyPackageFiles { get; set; }
 
+        /// <summary>
+        ///     Gets or sets a timeout in milliseconds for a request to sources
+        /// </summary>
+        public int RequestTimeout { get; set; } = defaultRequestTimeout;
```
JoC0de, line 68:
```suggestion
        public TimeSpan RequestTimeout { get; set; } = defaultRequestTimeout;
```
I think using `TimeSpan` is much easier to use as the string representation / parsing is usable without the user needing to calculate 'milliseconds'.

src/NuGetForUnity/Editor/NugetConfigFile.cs
```diff
@@ -62,6 +62,21 @@ public class NugetConfigFile
...
+
+        /// <summary>
+        ///     Default timeout in milliseconds for a request to sources
+        /// </summary>
+        private const int defaultRequestTimeout = 10000;
+
+        /// <summary>
+        ///     Gets or sets a value indicating whether the installed packages should be fixed
```
JoC0de, line 76:
```suggestion
        ///     Gets or sets a value indicating whether the installed packages should be fixed, so only the packages that are configure inside the 'package.config' are installed without installing the dependencies of them.
```

JoC0de (changes requested):
I just was busy. So here my review. It would be nice if you also add the new feature to the documentation (`Readme.md`).

src/NuGetForUnity/Editor/NugetConfigFile.cs
tete1030: I'm not sure about this. According to [TimeSpan](https://learn.microsoft.com/en-us/dotnet/api/system.timespan.parse?view=net-8.0), it accept strings like `00:00:10`. Requesting timeout is mostly below one minute I believe. Such timeout string is too much more expressive than needed. Maybe use 'seconds' instead of 'milliseconds' in both `Preferences` and xml config?

tete1030:
Thanks for your review. I will update these and the doc. Regarding `TimeSpan` I have a small question to discuss. Please see the comment I replied

src/NuGetForUnity/Editor/NugetConfigFile.cs
JoC0de: I'm not sure if the normal C# / unity developer has the knowledge of the string format of `TimeSpan` but it supports even nanoseconds precision with strings like `12:14:45.3448`. Yes a default like `00:01:30` so one and a half minute could eventually be misunderstood as 1 and a half hour. The alternative would be to use seconds but than we need to name the field `RequestTimeoutSeconds` so it is clear to the user what's the 'unit'. What do you think?

ConroyG:
Can I bump this?
I'm running into the same issue where a dependant package keeps comming back after I delete it and remove the packages.config reference

JoC0de:
@ConroyG OK if there is more demand I will try to fix the last small thinks and merge the feature on the weekend

JoC0de (approved):

tete1030:
Sorry that I forgot to follow through this PR. Really appreciate the merge.

PR #541, opened 2023-07-20
Restore improvements

popara96:
Fixed installing older versions of package when reloading from older package.config and added automatic restore when packages.config changes. Also fixed updating packages.


JoC0de (changes requested):

src/NuGetForUnity/Editor/NugetHelper.cs
```diff
@@ -1362,7 +1388,6 @@ public static bool Install(NugetPackage package, bool refreshAssets = true)
             {
                 WarnIfDotNetAuthenticationIssue(e);
                 Debug.LogErrorFormat("Unable to install package {0} {1}\n{2}", package.Id, package.Version, e);
-                installSuccess = false;
```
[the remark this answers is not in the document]
popara96: It is but it was initialized to `false` so Rider suggested either not initializing it or not reassigning the same value (`false`). I can do either of those, this just seemed more clean than changing `var installSuccess = false;` to `bool installSuccess;`.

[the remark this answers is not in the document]
JoC0de: I think removing the variable and doing `return true/false` is much more readable.

src/NuGetForUnity/Editor/NugetPackageAssetPostprocessor.cs
```diff
@@ -67,6 +67,20 @@ private void OnPreprocessAsset()
             LogResults(results);
         }
 
+        private static void OnPostprocessAllAssets(string[] importedAssets,
+            string[] deletedAssets,
+            string[] movedAssets,
+            string[] movedFromAssetPaths)
+        {
+            if (Array.IndexOf(importedAssets, NugetHelper.PackagesConfigFilePath) < 0)
```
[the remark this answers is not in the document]
popara96: Fixed this by expanding both paths to full before checking. The `PackagesConfigFilePath` will be moved to NuGetConfig in another PR after this is merged.
Maybe I am missing something but I am not sure how to Unit Test this method...but you were right, it was not working properly initially.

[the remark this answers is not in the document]
JoC0de: I prefer using absolute path as it can't be misused e.g. wenn you parse it to a different tool. Also it generates better exceptions e.g. when you get a DirectoryNotFoundException it is much easier to understand what is happening when the path is absolut.

src/NuGetForUnity/Editor/NugetHelper.cs
```diff
@@ -131,6 +131,11 @@ public static PackagesConfigFile PackagesConfigFile
             }
         }
 
+        public static void ReloadPackagesConfig()
```
JoC0de, line 134:
```suggestion
        /// <summary>
        ///     Invalidates the currently loaded 'package.config' so it is reloaded when it is accessed the next time.
        /// </summary>
        internal static void ReloadPackagesConfig()
```

JoC0de (changes requested):

src/NuGetForUnity.Tests/Assets/Tests/Editor/NuGetTests.cs
```diff
@@ -568,4 +569,116 @@ public void TestSerializeNugetPackageIdentifier(string version)
             Assert.That(deserialized.InRange(identifier), Is.True);
         }
     }
+
+    [Test]
+    [TestCase("jQuery", "3.7.0")]
+    public void TestPostprocessInstall(string packageId, string packageVersion)
+    {
+        var packagesFile = new XDocument();
```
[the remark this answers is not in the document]
popara96: I tried the first one but the `Packages` property has a private setter so I couldn't initialize it to empty list from here. So I changed it to the latter proposed, where I directly write what's going to be in the file, and I also see it as a more readable solution in context of a test.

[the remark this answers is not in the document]
popara96: Fixed it in a new commit to be able to create `PackagesConfigFile` in tests and use it to add packages.

src/NuGetForUnity.Tests/Assets/Tests/Editor/NuGetTests.cs
```diff
@@ -568,4 +569,63 @@ public void TestSerializeNugetPackageIdentifier(string version)
...
+        var package = new NugetPackageIdentifier(packageId, packageVersion) { IsManuallyInstalled = true };
+        var filepath = NugetHelper.PackagesConfigFilePath;
+
+        var packagesConfigFile = new PackagesConfigFile();
+        packagesConfigFile.AddPackage(package);
+        packagesConfigFile.Save(filepath);
+
+        Assert.IsFalse(NugetHelper.IsInstalled(package), "The package IS installed: {0} {1}", package.Id, package.Version);
+
+        AssetDatabase.Refresh();
```
JoC0de, line 586:
```suggestion
        NugetHelper.ReloadPackagesConfig();
        AssetDatabase.Refresh();
```

src/NuGetForUnity.Tests/Assets/Tests/Editor/NuGetTests.cs
```diff
@@ -568,4 +569,63 @@ public void TestSerializeNugetPackageIdentifier(string version)
...
+        var package = new NugetPackageIdentifier(packageId, packageVersion) { IsManuallyInstalled = true };
+        var filepath = NugetHelper.PackagesConfigFilePath;
+
+        NugetHelper.InstallIdentifier(package);
+        Assert.IsTrue(NugetHelper.IsInstalled(package), "The package was NOT installed: {0} {1}", package.Id, package.Version);
+
+        var packagesConfigFile = new PackagesConfigFile();
+        packagesConfigFile.Save(filepath);
+
+        AssetDatabase.Refresh();
```
JoC0de, line 604:
```suggestion
        NugetHelper.ReloadPackagesConfig();
        AssetDatabase.Refresh();
```

src/NuGetForUnity.Tests/Assets/Tests/Editor/NuGetTests.cs
```diff
@@ -568,4 +569,63 @@ public void TestSerializeNugetPackageIdentifier(string version)
...
+        var filepath = NugetHelper.PackagesConfigFilePath;
+
+        NugetHelper.InstallIdentifier(packageOld);
+        Assert.IsTrue(NugetHelper.IsInstalled(packageOld), "The package was NOT installed: {0} {1}", packageOld.Id, packageOld.Version);
+
+        var packagesConfigFile = new PackagesConfigFile();
+        packagesConfigFile.AddPackage(packageNew);
+        packagesConfigFile.Save(filepath);
+
+        AssetDatabase.Refresh();
```
JoC0de, line 625:
```suggestion
        NugetHelper.ReloadPackagesConfig();
        AssetDatabase.Refresh();
```

src/NuGetForUnity.Tests/Assets/Tests/Editor/NuGetTests.cs
```diff
@@ -568,4 +569,63 @@ public void TestSerializeNugetPackageIdentifier(string version)
             Assert.That(deserialized.InRange(identifier), Is.True);
         }
     }
+
```
JoC0de, line 572:
```suggestion
[TearDown]
public void CleanUp()
{
   NugetHelper.UninstallAll();
   if (File.Exists(NugetHelper.PackagesConfigFilePath))
   {
     File.Delete(NugetHelper.PackagesConfigFilePath);
   }
   NugetHelper.ReloadPackagesConfig();
}
```
I think this would make it easier so we don't need to cleanup in each Test.

JoC0de (changes requested):

src/NuGetForUnity.Tests/Assets/Tests/Editor/NuGetTests.cs
igor84: We already have CleanUp method at the top of the file that does UninstallAll and that method updates the PackagesConfig so there is no need to call Reload additionally.

src/NuGetForUnity.Tests/Assets/Tests/Editor/NuGetTests.cs
igor84: We do not want to call ReloadPackagesConfig() here since we want to test if AssetDatabase.Refresh properly does it by calling OnPostProcessAllAssets(), right?

src/NuGetForUnity.Tests/Assets/Tests/Editor/NuGetTests.cs
JoC0de: Ok then we can ignore this.

src/NuGetForUnity.Tests/Assets/Tests/Editor/NuGetTests.cs
JoC0de: Oh you are right. But why hare the tests than failing.

popara96: Not sure why they're failing on linux. I ran them locally on windows, Unity 2018.4.30f1, and they all passed...also the previous versions of this PR have been failing on the same step but not by failing tests, but by timing out if I remember correctly (after 360 minutes or so)

popara96:
The timeouts that happen seem to be related to [this bug](https://issuetracker.unity3d.com/issues/linux-platformimageforiconforextension-editor-log-spam-when-closing-the-editor-after-importing-a-project). I'm not sure what we can do with it except update Unity version, but we also modified the tests a bit in hopes to fix them.

JoC0de (changes requested):

JoC0de:
> The timeouts that happen seem to be related to [this bug](https://issuetracker.unity3d.com/issues/linux-platformimageforiconforextension-editor-log-spam-when-closing-the-editor-after-importing-a-project). I'm not sure what we can do with it except update Unity version, but we also modified the tests a bit in hopes to fix them.

It is strange that it only happens size I added the Local Package Hierarchy restore tests. I didn't update the unity version used in the Action.

src/NuGetForUnity/Editor/NugetPackageAssetPostprocessor.cs
```diff
@@ -67,6 +67,28 @@ private void OnPreprocessAsset()
             LogResults(results);
         }
 
+        /// <summary>
+        ///     Called when the asset database finishes importing assets.
+        ///     We use it to check if packages.config has been changed and if so, we want to restore packages.
+        /// </summary>
+        public static void OnPostprocessAllAssets(string[] importedAssets,
```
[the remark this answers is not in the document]
JoC0de: https://github.com/GlitchEnzo/NuGetForUnity/blob/653b9d58b0685bdd714e0d6f216b26a126f7a982/src/NuGetForUnity/Editor/NugetHelper.cs#L15

popara96:
@JoC0de we don't know for sure what causes this issue as it doesn't happen every time, but it seems that neither your nor my changes have anything to do with it. As for Unity 2018, the link I sent says that it is fixed in version 2018.4.35f1 (we're currently on 2018.4.30f1), so we might consider upgrading to at least that one, but it still does not guarantee the issue will disappear.

Issue #527, opened 2023-06-22
Feature Request: Move the NuGet and packages .dll into a folder (i.e. packages) make it dynamically findable?


Avalin:
Would be nice if it didn't clutter in the Assets folder itself :)

JoC0de:
Hi @Avalin thanks for the suggestion. Unfortunately it is a hard requirement from unity that the .dll's are placed somewhere inside the assets folder, else unity will just ignore them. You can configure the exact folder where they are placed.

PR #551, opened 2023-08-04
Made packages.config path configurable from NuGet preferences

popara96:
Minor changes in order to make `packages.config` path configurable. We don't allow for the path to be outside of `Assets` folder since we still need `OnPostprocessAllAssets` to be triggered if this file is changed.

### DIFF
--- a/src/NuGetForUnity.Tests/Assets/NuGet.config
+++ b/src/NuGetForUnity.Tests/Assets/NuGet.config
@@ -11,7 +11,7 @@
   </activePackageSource>
   <config>
     <add key="repositoryPath" value="./Packages" />
-    <add key="PackagesConfigPath" value="." />
+    <add key="PackagesConfigDirectoryPath" value="." />
     <add key="DefaultPushSource" value="http://www.nuget.org/api/v2/" />
     <add key="verbose" value="true" />
     <add key="ReadOnlyPackageFiles" value="false" />
--- a/src/NuGetForUnity.Tests/Assets/NuGet.config
+++ b/src/NuGetForUnity.Tests/Assets/NuGet.config
@@ -4,15 +4,16 @@
     <add key="NuGet" value="http://www.nuget.org/api/v2/" />
     <add key="V3" value="https://api.nuget.org/v3/index.json" />
   </packageSources>
+  <disabledPackageSources />
   <packageSourceCredentials />
   <activePackageSource>
     <add key="All" value="(Aggregate source)" />
   </activePackageSource>
   <config>
     <add key="repositoryPath" value="./Packages" />
+    <add key="PackagesConfigPath" value="." />
     <add key="DefaultPushSource" value="http://www.nuget.org/api/v2/" />
     <add key="verbose" value="true" />
-    <add key="PackagesConfigPath" value=".." />
     <add key="ReadOnlyPackageFiles" value="false" />
   </config>
 </configuration>
--- a/src/NuGetForUnity.Tests/Assets/NuGet.config
+++ b/src/NuGetForUnity.Tests/Assets/NuGet.config
@@ -1,4 +1,4 @@
-﻿<?xml version="1.0" encoding="utf-8" ?>
+﻿<?xml version="1.0" encoding="utf-8"?>
 <configuration>
   <packageSources>
     <add key="NuGet" value="http://www.nuget.org/api/v2/" />
@@ -12,6 +12,7 @@
     <add key="repositoryPath" value="./Packages" />
     <add key="DefaultPushSource" value="http://www.nuget.org/api/v2/" />
     <add key="verbose" value="true" />
+    <add key="PackagesConfigPath" value="Assets" />
     <add key="ReadOnlyPackageFiles" value="false" />
   </config>
 </configuration>
--- a/src/NuGetForUnity.Tests/Assets/NuGet.config
+++ b/src/NuGetForUnity.Tests/Assets/NuGet.config
@@ -12,7 +12,7 @@
     <add key="repositoryPath" value="./Packages" />
     <add key="DefaultPushSource" value="http://www.nuget.org/api/v2/" />
     <add key="verbose" value="true" />
-    <add key="PackagesConfigPath" value="Assets" />
+    <add key="PackagesConfigPath" value=".." />
     <add key="ReadOnlyPackageFiles" value="false" />
   </config>
 </configuration>
--- a/src/NuGetForUnity/Editor/NugetConfigFile.cs
+++ b/src/NuGetForUnity/Editor/NugetConfigFile.cs
@@ -74,30 +74,13 @@ namespace NugetForUnity
         /// </summary>
         public bool ReadOnlyPackageFiles { get; set; }
 
-        public string PackagesConfigPath;
+        public string PackagesConfigPath { get; set; }
 
         public string RelativePackagesConfigPath
         {
-            get => NugetHelper.GetProjectRelativePath(PackagesConfigPath);
+            get => NugetHelper.GetAssetsRelativePath(PackagesConfigPath);
 
-            private set
-            {
-                var tempValue = Environment.ExpandEnvironmentVariables(value);
-                if (tempValue == "Assets" || tempValue == "Assets/" || tempValue == "Assets\\")
-                {
-                    PackagesConfigPath = Application.dataPath;
-                }
-                else if (tempValue.StartsWith("Assets/") || tempValue.StartsWith("Assets\\"))
-                {
-                    PackagesConfigPath = Path.Combine(Application.dataPath, tempValue.Substring("Assets/".Length));
-                }
-                else
-                {
-                    Debug.LogWarning($"Path to packages.config in NuGet.config must be relative to the project root (attempted path: {tempValue}). Setting to default.");
-                    PackagesConfigPath = Application.dataPath;
-                    Save(NugetHelper.NugetConfigFilePath);
-                }
-            }
+            private set => PackagesConfigPath = Path.GetFullPath(Path.Combine(Application.dataPath, value));
         }
 
         /// <summary>
@@ -269,7 +252,7 @@ namespace NugetForUnity
             configFile.PackageSources = new List<INugetPackageSource>();
             configFile.InstallFromCache = true;
             configFile.ReadOnlyPackageFiles = false;
-            configFile.RelativePackagesConfigPath = "Assets";
+            configFile.RelativePackagesConfigPath = ".";
 
             var file = XDocument.Load(filePath);
 
@@ -421,7 +404,7 @@ namespace NugetForUnity
   </activePackageSource>
   <config>
     <add key=""repositoryPath"" value=""./Packages"" />
-    <add key=""PackagesConfigPath"" value=""Assets"" />
+    <add key=""PackagesConfigPath"" value=""."" />
   </config>
 </configuration>";
 
--- a/src/NuGetForUnity/Editor/NugetConfigFile.cs
+++ b/src/NuGetForUnity/Editor/NugetConfigFile.cs
@@ -74,6 +74,35 @@ namespace NugetForUnity
         /// </summary>
         public bool ReadOnlyPackageFiles { get; set; }
 
+        public string PackagesConfigPath;
+
+        public string RelativePackagesConfigPath
+        {
+            get => NugetHelper.GetProjectRelativePath(PackagesConfigPath);
+
+            private set
+            {
+                PackagesConfigPath = Path.GetFullPath(Environment.ExpandEnvironmentVariables(value));
+
+                var packageConfigRelativePath = NugetHelper.GetProjectRelativePath(PackagesConfigPath);
+
+                // package config path is invalid (not under Assets), set to default
+                if (packageConfigRelativePath == PackagesConfigPath || packageConfigRelativePath == ".")
+                {
+                    Debug.LogError("Path to packages.config in NuGet.config must be relative to the project root. Setting to default.");
+                    PackagesConfigPath = Application.dataPath;
+                    Save(NugetHelper.NugetConfigFilePath);
+                }
+
+                PackagesConfigPath = PackagesConfigPath.TrimEnd(Path.DirectorySeparatorChar, Path.AltDirectorySeparatorChar);
+            }
+        }
+
+        /// <summary>
+        ///     The path to the packages.config file.
+        /// </summary>
+        public string PackagesConfigFilePath => Path.Combine(PackagesConfigPath, PackagesConfigFile.FileName);
+
         /// <summary>
         ///     Gets or sets the timeout in seconds used for all web requests to NuGet sources.
         /// </summary>
@@ -145,6 +174,11 @@ namespace NugetForUnity
             addElement = new XElement("add");
             addElement.Add(new XAttribute("key", "repositoryPath"));
             addElement.Add(new XAttribute("value", savedRepositoryPath));
+            config.Add(addElement);
+
+            addElement = new XElement("add");
+            addElement.Add(new XAttribute("key", "PackagesConfigPath"));
+            addElement.Add(new XAttribute("value", RelativePackagesConfigPath));
             config.Add(addElement);
 
             // save the default push source
@@ -233,6 +267,7 @@ namespace NugetForUnity
             configFile.PackageSources = new List<INugetPackageSource>();
             configFile.InstallFromCache = true;
             configFile.ReadOnlyPackageFiles = false;
+            configFile.RelativePackagesConfigPath = "Assets";
 
             var file = XDocument.Load(filePath);
 
@@ -355,6 +390,10 @@ namespace NugetForUnity
                     {
                         configFile.LockPackagesOnRestore = bool.Parse(value);
                     }
+                    else if (string.Equals(key, "PackagesConfigPath", StringComparison.OrdinalIgnoreCase))
+                    {
+                        configFile.RelativePackagesConfigPath = value;
+                    }
                 }
             }
 
@@ -380,6 +419,7 @@ namespace NugetForUnity
   </activePackageSource>
   <config>
     <add key=""repositoryPath"" value=""./Packages"" />
+    <add key=""PackagesConfigPath"" value=""Assets"" />
   </config>
 </configuration>";
 
--- a/src/NuGetForUnity/Editor/NugetConfigFile.cs
+++ b/src/NuGetForUnity/Editor/NugetConfigFile.cs
@@ -82,18 +82,21 @@ namespace NugetForUnity
 
             private set
             {
-                PackagesConfigPath = Path.GetFullPath(Environment.ExpandEnvironmentVariables(value));
-
-                var packageConfigRelativePath = NugetHelper.GetProjectRelativePath(PackagesConfigPath);
-
-                // package config path is invalid (not under Assets), set to default
-                if (packageConfigRelativePath == PackagesConfigPath || packageConfigRelativePath == ".")
+                var tempValue = Environment.ExpandEnvironmentVariables(value);
+                if (tempValue == "Assets" || tempValue == "Assets/" || tempValue == "Assets\\")
                 {
-                    Debug.LogError("Path to packages.config in NuGet.config must be relative to the project root. Setting to default.");
                     PackagesConfigPath = Application.dataPath;
                 }
-
-                PackagesConfigPath = PackagesConfigPath.TrimEnd(Path.DirectorySeparatorChar, Path.AltDirectorySeparatorChar);
+                else if (tempValue.StartsWith("Assets/") || tempValue.StartsWith("Assets\\"))
+                {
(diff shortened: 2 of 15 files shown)


src/NuGetForUnity/Editor/NugetPreferences.cs
```diff
@@ -72,6 +74,34 @@ public override void OnGUI(string searchContext)
                 NugetHelper.NugetConfigFile.Verbose = verbose;
             }
 
+            EditorGUILayout.BeginHorizontal();
+            {
+                var packagesConfigPath = NugetHelper.NugetConfigFile.RelativePackagesConfigPath;
+                EditorGUILayout.LabelField($"Packages Config path: {packagesConfigPath}");
+                packagesConfigPath = Path.GetFullPath(packagesConfigPath);
```
igor84, line 81: Move this line inside the if statement bellow since it is only needed there.

igor84 (changes requested):

src/NuGetForUnity/Editor/NugetConfigFile.cs
```diff
@@ -74,6 +74,37 @@ public class NugetConfigFile
         /// </summary>
         public bool ReadOnlyPackageFiles { get; set; }
 
+        public string PackagesConfigPath;
```
JoC0de, line 77:
```suggestion
        public string PackagesConfigPath { get; set; }
```
Using public fields is a bad practice, use properties instead.

src/NuGetForUnity/Editor/NugetConfigFile.cs
```diff
@@ -74,6 +74,37 @@ public class NugetConfigFile
...
+            {
+                var tempValue = Environment.ExpandEnvironmentVariables(value);
+                if (tempValue == "Assets" || tempValue == "Assets/" || tempValue == "Assets\\")
+                {
+                    PackagesConfigPath = Application.dataPath;
+                }
+                else if (tempValue.StartsWith("Assets/") || tempValue.StartsWith("Assets\\"))
+                {
+                    PackagesConfigPath = Path.Combine(Application.dataPath, tempValue.Substring("Assets/".Length));
+                }
```
JoC0de, line 93:
```suggestion
                PackagesConfigPath = Path.Combine(NugetHelper.AbsoluteProjectPath, tempValue);
```

src/NuGetForUnity/Editor/PathHelper.cs
```diff
@@ -0,0 +1,223 @@
...
+            StringComparison.OrdinalIgnoreCase :
+            StringComparison.Ordinal;
+
+    /// <summary>
+    ///     Create a relative path from one path to another. Paths will be resolved before calculating the difference.
+    ///     Default path comparison for the active platform will be used (OrdinalIgnoreCase for Windows or Mac, Ordinal for Unix).
+    /// </summary>
+    /// <remarks>
+    ///     Should the project upgrade to Unity version that has .NET Standard 2.1, this can be removed and instead we will use:
+    ///     Path.GetRelativePath(relativeTo, path).
```
igor84, line 26: We actually can't use Path.GetRelativePath even then since it will not work correctly when calling the CLI version with a project path passed as argument that is different from current directory.

src/NuGetForUnity/Editor/NugetConfigFile.cs
```diff
@@ -147,6 +178,11 @@ public void Save(string filePath)
             addElement.Add(new XAttribute("value", savedRepositoryPath));
             config.Add(addElement);
 
+            addElement = new XElement("add");
+            addElement.Add(new XAttribute("key", "PackagesConfigPath"));
+            addElement.Add(new XAttribute("value", RelativePackagesConfigPath));
```
JoC0de, line 183: Should we really store the path even if it is just the default path?

src/NuGetForUnity/Editor/PathHelper.cs
```diff
@@ -0,0 +1,223 @@
...
+        {
+            throw new ArgumentNullException(nameof(relativeTo));
+        }
+
+        if (string.IsNullOrEmpty(path))
+        {
+            return path;
+        }
+
+        relativeTo = Path.GetFullPath(relativeTo);
```
igor84, line 46: We also can not use Path.GetFullPath since it will not return what we need when calling through CLI version with a path to the project passed as argument.

src/NuGetForUnity/Editor/NugetConfigFile.cs
```diff
@@ -74,6 +74,37 @@ public class NugetConfigFile
...
 
+        public string PackagesConfigPath;
+
+        public string RelativePackagesConfigPath
+        {
+            get => NugetHelper.GetProjectRelativePath(PackagesConfigPath);
+
+            private set
+            {
+                var tempValue = Environment.ExpandEnvironmentVariables(value);
```
JoC0de, line 85: If we support EnvironmentVariables inside the path. We need to store the "non expanded value" as loading and saving the NuGet.config file shouldn't change the value.

src/NuGetForUnity/Editor/PathHelper.cs
```diff
@@ -0,0 +1,223 @@
...
+        path = Path.GetFullPath(path);
```
igor84, line 47: This function should probably only accept full paths and not call GetFullPath on them. If someone passed relative paths to it this would not work correctly if called from CLI version since CurrentDirectory and project path set through the argument could be different. We could throw or return path if Path.IsPathRooted returned false for any of them.

src/NuGetForUnity.Tests/Assets/NuGet.config
```diff
@@ -12,6 +12,7 @@
     <add key="repositoryPath" value="./Packages" />
     <add key="DefaultPushSource" value="http://www.nuget.org/api/v2/" />
     <add key="verbose" value="true" />
+    <add key="PackagesConfigPath" value="Assets" />
```
JoC0de, line 15: When we store a path inside a file the "expected" behavior is that the path is relative to the file that contains the configuration. For `repositoryPath` it is the same. So the value `.` means the `package.config` file is placed in the same folder as the `NuGet.config` file.

src/NuGetForUnity/Editor/NugetPreferences.cs
```diff
@@ -72,6 +74,34 @@ public override void OnGUI(string searchContext)
...
+                if (GUILayout.Button("Browse"))
+                {
+                    var newPath = EditorUtility.OpenFolderPanel("Select Folder", packagesConfigPath, "");
+
+                    if (!string.IsNullOrEmpty(newPath) && newPath != packagesConfigPath)
+                    {
+                        var pathCheck = NugetHelper.GetProjectRelativePath(newPath);
+
+                        // make sure the path is within Assets directory
+                        if (!pathCheck.StartsWith("Assets"))
```
JoC0de, line 91: I think we can allow the file to be placed outside the "Assets" folder. I would only show a warning that this disables the functionality of restoring packages if the `package.config` file is changed "on the disk".

JoC0de (changes requested):

src/NuGetForUnity/Editor/PathHelper.cs
```diff
@@ -0,0 +1,223 @@
...
+    ///     Should the project upgrade to Unity version that has .NET Standard 2.1, this can be removed and instead we will use:
+    ///     Path.GetRelativePath(relativeTo, path).
+    ///     <see
+    ///         href="https://github.com/dotnet/runtime/blob/4aa77dfb0674fd61e41e574b8cc8ae63146f52c4/src/libraries/System.Private.CoreLib/src/System/IO/Path.cs#L843" />
+    ///     .
+    /// </remarks>
+    /// <param name="relativeTo">The source path the output should be relative to. This path is always considered to be a directory.</param>
+    /// <param name="path">The destination path.</param>
+    /// <returns>The relative path or <paramref name="path" /> if the paths don't share the same root.</returns>
+    internal static string GetRelativePath(string relativeTo, string path)
```
igor84, line 34: Why do we need this complex function or, in other words, why isn't GetProjectRelativePath we implemented more than enough for this project? If you just wanted to have a polyfill for Path.GetRelativePath that we can just remove when we can use the real thing I don't see that happening since as I wrote above I don't see how we could ever use Path.GetRelativePath and have it work as we would like when run from CLI version. Unless we make CLI version call Directory.SetCurrentDirectory on start to change current directory to the path passed as argument, but that might be strange...

src/NuGetForUnity/Editor/PathHelper.cs
JoC0de: The CLI can use `Directory.SetCurrentDirectory(projectPath);` to get the same behavior.

src/NuGetForUnity/Editor/PathHelper.cs
JoC0de: see other comment

src/NuGetForUnity/Editor/NugetConfigFile.cs
igor84: Should we support env variables for this? Seems weird that you depend on different packages in different environments considering how packages are usually made.

src/NuGetForUnity/Editor/PathHelper.cs
JoC0de: The difference between the "Easy" implementation you had before and the complex implementation (the polyfill of `GetProjectRelativePath`) is the support of having a `path` that is outside the `relativeTo` path. So e.g. the project relative path of `Assets/../../' is '..'. General example `C:/Test/Packages` relative to  `C:/Test/Assets` is  `../Packages`.

src/NuGetForUnity/Editor/NugetConfigFile.cs
JoC0de: Normally the `package.config` should live in the same GIT as the source code therefore relative path's should be enough to handle all scenarios.

src/NuGetForUnity/Editor/PathHelper.cs
JoC0de: There are some feature requests like #527 that like to have the `package.config` and the `NuGet.config` outside the `Assets` directory. And to support this we need relative path support that is not only going "down" the directory tree.

JoC0de:
@ popara96 i will fix the CLI soon.

src/NuGetForUnity/Editor/NugetConfigFile.cs
popara96: I think it would do no harm... It's changed to `.` now so it indicates the folder where `NuGet.config` is.

src/NuGetForUnity.Tests/Assets/NuGet.config
popara96: Fixed the behaviour now with this.

src/NuGetForUnity/Editor/PackagesConfigFile.cs
```diff
@@ -215,48 +215,57 @@ public void Save()
...
+        ///     path, makes a default version of the file on the new path.
+        /// </summary>
+        /// <param name="newPath">Path to move packages.config file to.</param>
         internal static void Move(string newPath)
         {
             var oldFilePath = NugetHelper.NugetConfigFile.PackagesConfigFilePath;
             var oldPath = NugetHelper.NugetConfigFile.PackagesConfigPath;
             NugetHelper.NugetConfigFile.PackagesConfigPath = newPath;
-            if (!File.Exists(oldFilePath))
+            try
```
popara96, line 228: Added try-catch in case there's any restriction for moving files (for example unauthorized access exception), in which case we log the error and set package config path to the old one.

src/NuGetForUnity/Editor/NugetConfigFile.cs
```diff
@@ -74,6 +74,20 @@ public class NugetConfigFile
         /// </summary>
         public bool ReadOnlyPackageFiles { get; set; }
 
+        public string PackagesConfigPath { get; set; }
```
JoC0de, line 77: Please add documentation

JoC0de, line 77:
```suggestion
        public string PackagesConfigDirectoryPath { get; set; }
```
I like to be explicit even if the name is long

src/NuGetForUnity/Editor/NugetConfigFile.cs
```diff
@@ -74,6 +74,20 @@ public class NugetConfigFile
...
+
+        public string RelativePackagesConfigPath
+        {
+            get => NugetHelper.GetAssetsRelativePath(PackagesConfigPath);
```
JoC0de, line 81: I think we should use `PathHelper.GetRelativePath` directly. This makes it clearer what happens. So the setter and the getter do the "same thing"

src/NuGetForUnity/Editor/NugetConfigFile.cs
```diff
@@ -147,6 +161,11 @@ public void Save(string filePath)
             addElement.Add(new XAttribute("value", savedRepositoryPath));
             config.Add(addElement);
 
+            addElement = new XElement("add");
+            addElement.Add(new XAttribute("key", "PackagesConfigPath"));
```
JoC0de, line 165: I would move `"PackagesConfigPath"` to a constant so we have no magic value, I did it for all new settings I added

src/NuGetForUnity/Editor/NugetPreferences.cs
```diff
@@ -72,6 +69,33 @@ public override void OnGUI(string searchContext)
...
+
+                    if (!string.IsNullOrEmpty(newPath) && newPath != packagesConfigPath)
+                    {
+                        var pathCheck = NugetHelper.GetAssetsRelativePath(newPath);
+
+                        // if the path root is different or it is not under Assets folder, we issue a warning
+                        if (pathCheck == newPath || pathCheck.StartsWith(".."))
+                        {
+                            Debug.LogWarning(
+                                "Putting packages.config outside of Assets folder disables the functionality of restoring packages if the file is changed on the disk.");
```
JoC0de, line 89: It would be nice if we show this in the UI as a warning under the setting.

JoC0de (changes requested):

src/NuGetForUnity/Editor/NugetConfigFile.cs
popara96: Fixed. I could do the same for the rest of the keys, I agree it would seem more clean that way.

src/NuGetForUnity/Editor/NugetPreferences.cs
```diff
@@ -82,19 +87,20 @@ public override void OnGUI(string searchContext)
...
                         var pathCheck = NugetHelper.GetAssetsRelativePath(newPath);
 
-                        // if the path root is different or it is not under Assets folder, we issue a warning
-                        if (pathCheck == newPath || pathCheck.StartsWith(".."))
-                        {
-                            Debug.LogWarning(
-                                "Putting packages.config outside of Assets folder disables the functionality of restoring packages if the file is changed on the disk.");
-                        }
+                        // if the path root is different or it is not under Assets folder, we want to show a warning message
+                        shouldShowPackagesConfigPathWarning = pathCheck == newPath || pathCheck.StartsWith("..");
```
igor84, line 91: If I change the path manually in the file and then open Preferences window this field will be false and I will not see the warning. You should probably initialize this field to a proper value in OnEnable().

src/NuGetForUnity/Editor/NugetPreferences.cs
```diff
@@ -20,6 +20,11 @@ public class NugetPreferences : SettingsProvider
         /// </summary>
         private static Vector2 scrollPosition;
 
+        /// <summary>
+        ///     Indicates if the warning for packages.config file path should be shown in case it is outside of Assets folder.
+        /// </summary>
+        private static bool shouldShowPackagesConfigPathWarning;
```
igor84, line 26: I don't see a reason for this and scrollPosition fields to be static. See if it works when you change them not to be static.

igor84 (approved):

src/NuGetForUnity/Editor/NugetPreferences.cs
JoC0de, line 91:
```suggestion
                        shouldShowPackagesConfigPathWarning = pathCheck.StartsWith("..");
```
I would always show the warning if the path is outside the assets folder

src/NuGetForUnity/Editor/NugetConfigFile.cs
```diff
@@ -74,6 +76,22 @@ public class NugetConfigFile
         /// </summary>
         public bool ReadOnlyPackageFiles { get; set; }
 
+        /// <summary>
+        ///     Gets or sets absolute path to directory containing packages.config file.
+        /// </summary>
+        public string PackagesConfigDirectoryPath { get; set; }
+
+        public string RelativePackagesConfigPath
```
JoC0de, line 84:
```suggestion
        /// <summary>
        ///     Gets or sets the relative path to directory containing packages.config file. The path is relative to the folder containing the 'NuGet.config' file.
        /// </summary>
        public string RelativePackagesConfigDirectoryPath
```

src/NuGetForUnity/Editor/NugetHelper.cs
```diff
@@ -125,6 +133,16 @@ public static PackagesConfigFile PackagesConfigFile
             }
         }
 
+        /// <summary>
+        ///     Returns the path relative to Assets directory, or <c>"."</c> if it is the Assets directory.
+        /// </summary>
+        /// <param name="path">The path of witch we calculate the relative path of.</param>
+        /// <returns>The path relative to Assets directory, or <c>"."</c> if it is the Assets directory.</returns>
+        public static string GetAssetsRelativePath(string path)
```
JoC0de, line 141:
```suggestion
        internal static string GetAssetsRelativePath(string path)
```

src/NuGetForUnity/Editor/NugetPreferences.cs
```diff
@@ -72,6 +74,34 @@ public override void OnGUI(string searchContext)
...
+                        PackagesConfigFile.Move(newPath);
+                        preferencesChangedThisFrame = true;
+                    }
+                }
+            }
+            EditorGUILayout.EndHorizontal();
+            if (shouldShowPackagesConfigPathWarning)
+            {
+                EditorGUILayout.HelpBox(
+                    "Putting packages.config outside of Assets folder disables the functionality of restoring packages if the file is changed on the disk.", MessageType.Warning);
```
JoC0de, line 102:
```suggestion
                    "The packages.config is placed outside the Assets folder, this disables the functionality of automatically restoring packages if the file is changed on the disk.", MessageType.Warning);
```
Something like this?

src/NuGetForUnity/Editor/PackagesConfigFile.cs
```diff
@@ -212,6 +215,59 @@ public void Save(string filePath)
...
+                    AssetDatabase.Refresh();
+                    return;
+                }
+
+                var newFilePath = Path.GetFullPath(Path.Combine(newPath, FileName));
+
+                Directory.CreateDirectory(newPath);
+
+                // moving config to the new path
+                File.Move(oldFilePath, newFilePath);
```
JoC0de, line 244:
```suggestion
                File.Move(oldFilePath, newFilePath);
                
            }
            catch (Exception e)
            {
                Debug.LogException(e);
                NugetHelper.NugetConfigFile.PackagesConfigDirectoryPath = oldPath;
            }
```
Resetting to `oldPath` only works if we didn't move the file successfully. I think this is not a problem when we ignore (not explicitly handle) errors occurring later as moving the .meta etc. can be considered "utility" as it is not essential for moving the file.

JoC0de (changes requested):
Sorry for that much changes. Together we are getting a nice solution. 👍

src/NuGetForUnity/Editor/NugetConfigFile.cs
popara96: Added this but without `sets` as Rider reminded mi the setter is private and therefore shouldn't be a part of the summary 😄

src/NuGetForUnity/Editor/NugetPreferences.cs
```diff
@@ -16,14 +17,23 @@ public class NugetPreferences : SettingsProvider
...
+        private bool shouldShowPackagesConfigPathWarning;
 
         /// <summary>
         ///     Initializes a new instance of the <see cref="NugetPreferences" /> class.
+        ///     Path of packages.config file is checked here as well in case it was manually changed.
         /// </summary>
         public NugetPreferences()
             : base("Preferences/NuGet For Unity", SettingsScope.User)
         {
+            var packagesConfigPath = NugetHelper.NugetConfigFile.PackagesConfigDirectoryPath;
```
JoC0de, line 34: I would move the check code to a function so we don't need it twice

JoC0de (changes requested):

JoC0de (approved):
Now we can merge 👍

popara96:
Thank you for the thorough review @JoC0de !

JoC0de:
Thank you for the improvement @popara96

PR #552, opened 2023-08-04
Minor improvement of menu item options

popara96:
Added separate 'Preferences' menu item and set 'Version' option to redirect to the currently installed release page


JoC0de (approved):
👍  Good idea i also found it strange that the version opens the preferences.

igor84 (approved):

PR #554, opened 2023-08-12
Package version selection

### LINKED ISSUE #409: package version selection
please add package version selection when installing package

JoC0de:
Allow selection what version of a package should be installed.
image

This also removes the "Show all Versions" checkbox because it has no actual usage.
- Packages fetched using Local or V3 API package source support the `Versions` list so the version can be selected using the drop down.
- Packages fetched using V2 API never supported these feature (at least not the nuget.org V2 API)

fixes #409

### DIFF
--- a/src/NuGetForUnity/Editor/NugetWindow.cs
+++ b/src/NuGetForUnity/Editor/NugetWindow.cs
@@ -114,16 +114,6 @@ namespace NugetForUnity
         [SerializeField]
         private List<SerializableNugetPackage> serializableUpdatePackages;
 
-        /// <summary>
-        ///     True to show all old package versions.  False to only show the latest version.
-        /// </summary>
-        private bool showAllOnlineVersions;
-
-        /// <summary>
-        ///     True to show all old package versions.  False to only show the latest version.
-        /// </summary>
-        private bool showAllUpdateVersions;
-
         private bool showImplicitlyInstalled;
 
         /// <summary>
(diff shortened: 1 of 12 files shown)


src/NuGetForUnity/Editor/NugetWindow.cs
```diff
@@ -1012,7 +1002,35 @@ private void DrawPackage(INugetPackage package, GUIStyle packageStyle, GUIStyle
...
+                        versionDropdownData.DropdownOptions = versionDropdownData.SortedVersions.Select(version => version.FullVersion).ToArray();
+                        versionDropdownData.SelectedIndex = versionDropdownData.SortedVersions.IndexOf(package.PackageVersion);
+                        versionDropdownDataPerPackage.Add(package.Id, versionDropdownData);
+                    }
+
+                    GUILayout.Label("Version");
+                    versionDropdownData.SelectedIndex = EditorGUILayout.Popup(
+                        versionDropdownData.SelectedIndex,
+                        versionDropdownData.DropdownOptions,
+                        GUILayout.Width(versionDropdownData.CalculatedMaxWith));
```
igor84, line 1031: What happens if you don't pass the Width explicitly since I would expect Popup to handle the width like this automatically?

igor84 (approved):
Everything looks good but are you sure ShowAllVersions didn't work with V2. I know I saw recently that it doesn't work on Install tab, but I remember seeing it work some time before. Maybe it was only working on Update tab...

JoC0de:
> Everything looks good but are you sure ShowAllVersions didn't work with V2. I know I saw recently that it doesn't work on Install tab, but I remember seeing it work some time before. Maybe it was only working on Update tab...

I never testet what happens if you check the ShowAllVersions on the Update tab. But keeping the option would mean that we need to decide if the V2 API is used because the option only makes sense for it. Is this a useful setting? You install a package and than you are able to upgrade / downgrade it on the update tab.

src/NuGetForUnity/Editor/NugetWindow.cs
JoC0de: If I do nothing it has a fixed "to big" width. I tried getting it to "schrink" using `GUILayout.ExpandWith(false)`, `GUILayout.MaxWith(5)` and even tried to use GuiStyle but unfortunately nothing worked.

igor84:
I almost never needed it and manually editing packages.json file on one or two occasions I did need it was fine with me. I guess the best course is to remove it like you did and only think about it if people start complaining :).

PR #562, opened 2023-09-07
Fix Invalid File Path Characters in Cached Image Hashing

robrab2000:
This pull request addresses an issue with the caching of images where the Base64-encoded hash could produce file paths with invalid characters. The issue was discovered while using an Apple Silicon based MacBook Pro running macOS Ventura (13.5.1) and Unity3d 2021.3.22f1.

Changes made:
- Modified the `GetHash` method to convert the MD5 hash from Base64 encoding to a hexadecimal string. This ensures the generated hash string does not contain invalid characters for file paths.
- Explicitly specified the default encoding in the `GetHash` method.

These changes resolve caching issues on disk and ensure compatibility across different operating systems.

### DIFF
--- a/src/NuGetForUnity/Editor/Helper/NugetPackageTextureHelper.cs
+++ b/src/NuGetForUnity/Editor/Helper/NugetPackageTextureHelper.cs
@@ -116,7 +116,15 @@ namespace NugetForUnity.Helper
 
             var md5 = new MD5CryptoServiceProvider();
             var data = md5.ComputeHash(Encoding.Default.GetBytes(s));
-            return Convert.ToBase64String(data);
+
+            // Convert byte array to hexadecimal string
+            var sb = new StringBuilder();
+            for (int i = 0; i < data.Length; i++)
+            {
+                sb.Append(data[i].ToString("x2"));
+            }
+
+            return sb.ToString();
         }
     }
 }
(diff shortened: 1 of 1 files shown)


JoC0de:
Hi @robrab2000 
Thanks for the PR. I just changed the Hex to base 64 two weeks ago. I didn't expect that any of the characters are not allowed on some systems. Do you know what char it was?

robrab2000:
Hey, yeah it was the '/' that was giving troubles :)

src/NuGetForUnity/Editor/Helper/NugetPackageTextureHelper.cs
```diff
@@ -116,7 +116,15 @@ private static string GetHash(string s)
...
-            return Convert.ToBase64String(data);
+
+            // Convert byte array to hexadecimal string
+            var sb = new StringBuilder();
+            for (int i = 0; i < data.Length; i++)
+            {
+                sb.Append(data[i].ToString("x2"));
+            }
+
+            return sb.ToString();
```
JoC0de, line 127:
```suggestion
            return Convert.ToBase64String(data).Replace('+', '-').Replace('/', '_').Replace("=", null);
```
It would basically be Base64Url encoding. So only `[A-Za-z]`, `'-'` and `'_'`. Should be supported by all operating systems. And we will have a single line 😄

JoC0de (changes requested):

Issue #557, opened 2023-08-15
Stop decompressing files that are not needed to speed up installing


igor84:
Currently when we are installing a nupkg we decompress all the files from it and then we call the `NugetHelper.CleanInstallationDirectory()` method that deletes a number of files that are not needed in Unity. We should instead skip the decompression of files that are not needed which will noticeably speed up the installation.

PR #566, opened 2023-09-19
Optimizing installation duration

popara96:
Avoid deleting folders during clean after install by skipping the extraction of unneeded folders and files. [#557 ](https://github.com/GlitchEnzo/NuGetForUnity/issues/557)

### DIFF
--- a/src/NuGetForUnity/Editor/NugetPackageInstaller.cs
+++ b/src/NuGetForUnity/Editor/NugetPackageInstaller.cs
@@ -217,7 +217,7 @@ namespace NugetForUnity
                     // copy the .nupkg inside the Unity project
                     File.Copy(cachedPackagePath, Path.Combine(baseDirectory, package.PackageFileName), true);
 
-                    static void ExtractPackageEntry(ZipArchiveEntry entry, string baseDir)
+                    void ExtractPackageEntry(ZipArchiveEntry entry, string baseDir)
                     {
                         var filePath = Path.Combine(baseDir, entry.FullName);
                         var directory = Path.GetDirectoryName(filePath) ??
--- a/src/NuGetForUnity/Editor/NugetPackageInstaller.cs
+++ b/src/NuGetForUnity/Editor/NugetPackageInstaller.cs
@@ -183,60 +183,45 @@ namespace NugetForUnity
 
                             // we don't want to unpack all lib folders and then delete all but one; rather we will decide the best
                             // target framework before unpacking, but first we need to collect all lib entries from zip
-                            if (entryFullName.StartsWith("lib"))
+                            const string libDirectoryName = "lib/";
+                            if (entryFullName.StartsWith(libDirectoryName))
                             {
-                                const int frameworkStartIndex = 4; // length of "lib/"
-                                var secondSlashIndex = entryFullName.IndexOf("/", frameworkStartIndex, StringComparison.Ordinal);
-                                var framework = entryFullName.Substring(frameworkStartIndex, secondSlashIndex - frameworkStartIndex);
-                                if (libs.TryGetValue(framework, out var entryList))
+                                var frameworkStartIndex = libDirectoryName.Length;
+                                var secondSlashIndex = entryFullName.IndexOf('/', frameworkStartIndex);
+                                if (secondSlashIndex == -1)
                                 {
-                                    entryList.Add(entry);
+                                    // a file inside lib folder -> we skip it
+                                    continue;
                                 }
-                                else
+
+                                var framework = entryFullName.Substring(libDirectoryName.Length, secondSlashIndex - frameworkStartIndex);
+                                if (!libs.TryGetValue(framework, out var entryList))
                                 {
-                                    libs[framework] = new List<ZipArchiveEntry> { entry };
+                                    entryList = new List<ZipArchiveEntry>();
+                                    libs.Add(framework, entryList);
                                 }
+
+                                entryList.Add(entry);
 
                                 continue;
                             }
 
-                            ExtractPackageEntry(entry, baseDirectory);
+                            PackageContentManager.ExtractPackageEntry(entry, baseDirectory);
                         }
 
                         // go through all lib zip entries and find the best target framework, then unpack it
-                        var bestFramework = TargetFrameworkResolver.TryGetBestTargetFramework(libs.Keys, key => key);
-                        if (bestFramework != null)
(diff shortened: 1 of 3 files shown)


src/NuGetForUnity/Editor/NugetPackageInstaller.cs
```diff
@@ -228,5 +283,64 @@ private static bool Install(INugetPackage package, bool refreshAssets, bool inst
                 }
             }
         }
+
+        private static bool ShouldSkipUnpackingOnPath(string path, string packageId)
```
igor84, line 287: Maybe this method should also be in PackageContentManager class?

src/NuGetForUnity/Editor/NugetPackageInstaller.cs
```diff
@@ -170,9 +171,36 @@ private static bool Install(INugetPackage package, bool refreshAssets, bool inst
...
+                            }
+
+                            // we don't want to unpack all lib folders and then delete all but one; rather we will decide the best
+                            // target framework before unpacking, but first we need to collect all lib entries from zip
+                            if (entryFullName.StartsWith("lib"))
+                            {
+                                const int frameworkStartIndex = 4; // length of "lib/"
+                                var secondSlashIndex = entryFullName.IndexOf("/", frameworkStartIndex, StringComparison.Ordinal);
+                                var framework = entryFullName.Substring(frameworkStartIndex, secondSlashIndex - frameworkStartIndex);
+                                if (libs.ContainsKey(framework))
```
igor84, line 193: Maybe this can be written a bit better:
```c#
if (libs.TryGet(framework, out var entryList)
{
    entryList.Add(entry);
}
else
{
    libs[framework] = new List<ZipArchiveEntry> { entry };
}
```

src/NuGetForUnity/Editor/NugetPackageInstaller.cs
```diff
@@ -191,6 +219,33 @@ private static bool Install(INugetPackage package, bool refreshAssets, bool inst
...
+                                    continue;
+                                }
+
+                                entry.ExtractToFile(filePath, true);
+
+                                if (ConfigurationManager.NugetConfigFile.ReadOnlyPackageFiles)
+                                {
+                                    var extractedFile = new FileInfo(filePath);
+                                    extractedFile.Attributes |= FileAttributes.ReadOnly;
+                                }
```
igor84, line 246: Extract this to local method so it is not repeated here and above.

igor84 (changes requested):

igor84 (approved):

src/NuGetForUnity/Editor/PackageContentManager.cs
```diff
@@ -173,6 +98,65 @@ internal static void CleanInstallationDirectory(INugetPackageIdentifier package)
...
+            if (path.EndsWith($"{packageId}.nuspec.meta", StringComparison.Ordinal)) return true;
+
+            // skip directories & files that NuGet normally deletes
+            if (path.StartsWith("_rels", StringComparison.Ordinal) || path.Contains("/_rels/")) return true;
+            if (path.StartsWith("package", StringComparison.Ordinal) || path.Contains("/package/")) return true;
+            if (path.EndsWith($"{packageId}.nuspec", StringComparison.Ordinal)) return true;
+            if (path.EndsWith("[Content_Types].xml", StringComparison.Ordinal)) return true;
+
+            // Unity has no use for the build directory
+            if (path.StartsWith("build", StringComparison.Ordinal) || path.Contains("/build/")) return true;
```
JoC0de, line 113:
```suggestion
            if (path.StartsWith("build/", StringComparison.Ordinal) || path.Contains("/build/")) return true;
```
The `path` is always a file path so to match a directory name we need the `/`

src/NuGetForUnity/Editor/PackageContentManager.cs
```diff
@@ -173,6 +98,65 @@ internal static void CleanInstallationDirectory(INugetPackageIdentifier package)
...
+            if (path.StartsWith("ref", StringComparison.Ordinal) || path.Contains("/ref/")) return true;
+
+            // Skip all PDB files since Unity uses Mono and requires MDB files, which causes it to output "missing MDB" errors
+            if (path.EndsWith(".pdb", StringComparison.Ordinal)) return true;
+
+            // Skip all folders that contain localization resource file
+            // Format of these entries is lib/<framework>/<language-code>/...
+            if (path.StartsWith("lib", StringComparison.Ordinal) || path.Contains("/lib/"))
+            {
+                var libSlashIndex = path.IndexOf("lib/", StringComparison.Ordinal) + 4;
```
JoC0de, line 136:
```suggestion
            const string libDirectoryName = "lib/";
            if (path.StartsWith(libDirectoryName, StringComparison.Ordinal) || path.Contains("/lib/"))
            {
                var libSlashIndex = path.IndexOf(libDirectoryName, StringComparison.Ordinal) + libDirectoryName.Length;
```

JoC0de (changes requested):
First review, I didn't had much time ...

src/NuGetForUnity/Editor/NugetPackageInstaller.cs
```diff
@@ -170,31 +171,72 @@ private static bool Install(INugetPackage package, bool refreshAssets, bool inst
...
-                            entry.ExtractToFile(filePath, true);
+                            ExtractPackageEntry(entry, baseDirectory);
+                        }
 
-                            if (ConfigurationManager.NugetConfigFile.ReadOnlyPackageFiles)
+                        // go through all lib zip entries and find the best target framework, then unpack it
+                        var bestFramework = TargetFrameworkResolver.TryGetBestTargetFramework(libs.Keys, key => key);
+                        if (bestFramework != null)
+                        {
+                            foreach (var entry in libs[bestFramework])
```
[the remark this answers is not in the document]
popara96: There's a problem with this. I added a constraint to `TryGetBestTargetFramework` method that `T` has to be a class, because at first I tried what you suggested, but the default value for key-value pair is not a `null` value, rather an empty KVP ([,]), and we check for `null` there when asking for `bestMatch`. I could maybe instead ask something like `if (bestMatch != default(T))`?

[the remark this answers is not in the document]
JoC0de: I used `if (bestFrameworkEntries.Values != null)` (see in commit)

JoC0de (approved):

PR #570, opened 2023-09-26
Avoid unzipping .nupkg every time we need to read from .nuspec by saving .nuspec inside the package folder instead of .nupkg

popara96:
Backwards compatibility is managed by checking for .nupkg files and if there are some, we unpack .nuspec file from them and delete them. Tests have also been changed accordingly.


JoC0de:
Hi,
The failing tests are the one that try to install a nuget package from a lokal package source. Most probably the issue is in the setup because it needs the .nupkg file from a previous install to create a lokal package source.

src/NuGetForUnity/Editor/PackageSource/NugetPackageSourceLocal.cs
igor84, line 1: Since Local Source is a folder that contains nupkg files that you can then install in your project all changes in this file and NuspecFile.cs should be reverted to still work with nupkg.

igor84 (approved):

src/NuGetForUnity.Tests/Assets/Tests/Editor/NuGetTests.cs
```diff
@@ -317,21 +317,24 @@ public void InstallAndSearchLocalPackageSource([Values] bool hierarchical)
...
+            var nuspecFilePath = Path.Combine(
                 ConfigurationManager.NugetConfigFile.RepositoryPath,
                 $"{package.Id}.{package.Version}",
-                package.PackageFileName);
-            Assert.That(packageFilePath, Does.Exist.IgnoreDirectories);
+                package.SpecificationFileName);
+            Assert.That(nuspecFilePath, Does.Exist.IgnoreDirectories);
+
+            var nupkgFileName = $"{package.Id}.{package.Version}.nupkg";
+            var nupkgFilePath = Path.Combine(PackageCacheManager.CacheOutputDirectory, nupkgFileName);
```
JoC0de, line 328:
```suggestion
Assert.That(nupkgFilePath, Does.Exist.IgnoreDirectories);
```

src/NuGetForUnity/Editor/InstalledPackagesManager.cs
```diff
@@ -171,13 +172,32 @@ internal static void UpdateInstalledPackages()
             var manuallyInstalledPackagesNumber = 0;
 
             // a package that was installed via NuGet will have the .nupkg it came from inside the folder
+            // NOTE : this is useful only for backwards compatibility
```
JoC0de, line 175:
```suggestion
            // a package that was installed via a old version of NuGetForUnity will have the .nupkg it came from inside the folder, as we updated the behavior to only keep the .nuspec. So we need to extract the .nuspec from the .nupkg
```

src/NuGetForUnity/Editor/InstalledPackagesManager.cs
```diff
@@ -171,13 +172,32 @@ internal static void UpdateInstalledPackages()
...
+                // delete .nupkg and its .meta file to sync compatibility
+                File.Delete(nupkgFile);
+                var metaFile = $"{nupkgFile}.meta";
+                if (File.Exists(metaFile))
+                {
+                    File.Delete($"{nupkgFile}.meta");
+                }
             }
 
             // if the source code & assets for a package are pulled directly into the project (ex: via a symlink/junction) it should have a .nuspec defining the package
```
JoC0de, line 203: We should update the comment.

src/NuGetForUnity/Editor/InstalledPackagesManager.cs
```diff
@@ -226,16 +246,17 @@ internal static bool RemoveUnnecessaryPackages()
                     continue;
                 }
 
-                var pkgPath = Path.Combine(folder, $"{folderName}.nupkg");
-                if (!File.Exists(pkgPath))
+                Directory.GetFiles(folder, "*.nuspec");
```
JoC0de, line 249:
```suggestion
```

src/NuGetForUnity/Editor/InstalledPackagesManager.cs
```diff
@@ -226,16 +246,17 @@ internal static bool RemoveUnnecessaryPackages()
...
+                    // ignore folder not containing a nuspec file
                     continue;
                 }
 
-                var package = NugetPackageLocal.FromNupkgFile(
-                    pkgPath,
-                    new NugetPackageSourceLocal("Nupkg file already installed", Path.GetDirectoryName(pkgPath)));
+                var package = NugetPackageLocal.FromNuspecFile(
+                    nuspecPath,
+                    new NugetPackageSourceLocal("Nupkg file already installed", Path.GetDirectoryName(nuspecPath)));
```
JoC0de, line 259:
```suggestion
                    new NugetPackageSourceLocal("Nuspec file already installed", Path.GetDirectoryName(nuspecPath)));
```

src/NuGetForUnity/Editor/Models/NugetPackageV2.cs
```diff
@@ -54,5 +54,18 @@ public static NugetPackageV2 FromNupkgFile(string nupkgFilePath, NugetPackageSou
...
             return package;
         }
+
+        /// <summary>
+        ///     Loads a <see cref="NugetPackageV2" /> from the .nuspec file at the given file-path.
+        /// </summary>
+        /// <param name="nuspecFilePath">The file-path to the .nuspec file to load.</param>
+        /// <param name="packageSource">The source this package was downloaded with / provided by.</param>
+        /// <returns>The <see cref="NugetPackageV2" /> loaded from the .nuspec file.</returns>
+        public static NugetPackageV2 FromNuspecFile(string nuspecFilePath, NugetPackageSourceV2 packageSource)
```
JoC0de, line 64: I think this function is never used.

JoC0de (changes requested):

src/NuGetForUnity/Editor/Models/NugetPackageV2.cs
popara96: You're right, it's not, but the function above (`FromNupkgFile`) was not used as well and I wanted to make it similar to how it is in `NugetPackageLocal` :) I will delete this one as per comment, and I can remove the one above as well.

JoC0de: Yes remove both, the don't make sense as they need a remote package source, for a local file.

src/NuGetForUnity.Tests/Assets/Tests/Editor/NuGetTests.cs
JoC0de: I can't see the change in the current version

JoC0de (changes requested):
We need to rebase, and there is one missing conversation

src/NuGetForUnity.Tests/Assets/Tests/Editor/NuGetTests.cs
popara96: Not sure why, I added `Assert.That(nupkgFilePath, Does.Exist.IgnoreDirectories);` on line `328`, it's right after `var nupkgFilePath = Path.Combine(PackageCacheManager.CacheOutputDirectory, nupkgFileName);` if I understood the requested change correctly?

popara96:
After this gets approved, I will push the squashed commit directly to master.

src/NuGetForUnity.Tests/Assets/Tests/Editor/NuGetTests.cs
JoC0de: Am I blind:
https://github.com/GlitchEnzo/NuGetForUnity/blob/501fd3437ee5b46badadfe4c40dc8f0bda0a41ad/src/NuGetForUnity.Tests/Assets/Tests/Editor/NuGetTests.cs#L326-L329

JoC0de: Ok my fault

JoC0de (approved):
@popara96 why directly pushing to master. I don't like when PRs are skipped as they ensure e.g. that the CI works. I would just squash the commits -> rebase -> merge PR. So we keep the link to the PR / it will be listed correctly inside the Release Nodes.

igor84:
Does that mean we force push the branch after rebase + squash? We were not sure if we force push was allowed.

JoC0de:
> Does that mean we force push the branch after rebase + squash? We were not sure if we force push was allowed.

Yes force-push on branches is allowed. At leas I do it, but I always create branches on my fork. Without force-push it would be also possible but then you need to do a merge commit (I also prefer rebase over merging)

PR #576, opened 2023-10-02
Adding slim restore option 

popara96:
Option for skipping dependency installation and checking for already imported Unity libs on auto restore.

### DIFF
--- a/src/NuGetForUnity/Editor/Configuration/NugetConfigFile.cs
+++ b/src/NuGetForUnity/Editor/Configuration/NugetConfigFile.cs
@@ -44,8 +44,6 @@ namespace NugetForUnity.Configuration
 
         private const string RequestTimeoutSecondsConfigKey = "RequestTimeoutSeconds";
 
-        private const string LockPackagesOnRestoreConfigKey = "LockPackagesOnRestore";
-
         private const string PackagesConfigDirectoryPathConfigKey = "PackagesConfigDirectoryPath";
 
         /// <summary>
@@ -89,6 +87,12 @@ namespace NugetForUnity.Configuration
         public bool Verbose { get; set; }
 
         /// <summary>
+        ///     Gets or sets a value indicating whether to skip installing dependencies and checking for pre-imported Unity libs
+        ///     while auto-restoring.
+        /// </summary>
+        public bool SlimRestore { get; set; }
+
+        /// <summary>
         ///     Gets or sets a value indicating whether a package is installed from the cache (if present), or if it always downloads the package from the
         ///     server.
         /// </summary>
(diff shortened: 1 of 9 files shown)


igor84 (approved):

src/NuGetForUnity/Editor/Configuration/NugetConfigFile.cs
```diff
@@ -88,6 +86,12 @@ public class NugetConfigFile
         /// </summary>
         public bool Verbose { get; set; }
 
+        /// <summary>
+        ///     Gets or sets a value indicating whether to skip installing dependencies and checking for pre-imported Unity libs
+        ///     while auto-restoring.
+        /// </summary>
+        public bool SlimRestore { get; set; }
```
JoC0de, line 93:
```suggestion
        public bool SlimRestore { get; set; } = true;
```
To switch it on 'by default' if someone updates to the new version of NuGetForUnity. I think it is valid to enable it by default if we add a not inside the Release Notes. Or what do you think @igor84, @popara96?

src/NuGetForUnity/Editor/Configuration/NugetConfigFile.cs
```diff
@@ -402,6 +401,14 @@ public void Save([NotNull] string filePath)
                 config.Add(addElement);
             }
 
+            if (SlimRestore)
+            {
+                addElement = new XElement("add");
+                addElement.Add(new XAttribute("key", "slimRestore"));
+                addElement.Add(new XAttribute("value", Verbose.ToString().ToLowerInvariant()));
+                config.Add(addElement);
+            }
```
JoC0de, line 410:
```suggestion
            addElement = new XElement("add");
            addElement.Add(new XAttribute("key", "slimRestore"));
            addElement.Add(new XAttribute("value", SlimRestore.ToString().ToLowerInvariant()));
            config.Add(addElement);
```
Especially if we set `SlimRestore` to `true` by default I would always include the setting in the `NuGet.config` file.
Also there is a copy-paste error `Verbose` -> `SlimRestore`

src/NuGetForUnity/Editor/Ui/NugetPreferences.cs
```diff
@@ -92,6 +92,17 @@ public override void OnGUI([CanBeNull] string searchContext)
                 ConfigurationManager.NugetConfigFile.Verbose = verbose;
             }
 
+            var slimRestore = EditorGUILayout.Toggle(
+                new GUIContent(
+                    "Slim Restore",
+                    "When slim restore is enabled, installing dependencies and checking against Unity pre-imported libs will be skipped on automatic restore. Manually restoring via menu option will ignore this setting."),
```
JoC0de, line 98:
```suggestion
                    "Slim restore is a faster way of installing / restoring packages after opening the Unity project. To achieve this the package installation step skips installing dependencies not listed inside the package.config, also it skips checking against Unity pre-imported libraries. If you have a complex project setup with multiple target platforms that include different packages it could be that you need to disable this feature. Manually restoring via menu option will ignore this setting."),
```
Not perfect but I think it is more understandable for users not "into it", provide some information so he can decide if he should enable / disable the feature.

JoC0de (changes requested):

src/NuGetForUnity/Editor/Configuration/NugetConfigFile.cs
igor84: We thought about it. We are not sure what the best option is but we concluded the safest is to leave this false in existing projects so they continue working as they did and users having to manually enable it in old projects if they wish to change the behavior, while we default to true only for new projects.

src/NuGetForUnity/Editor/Ui/NugetPreferences.cs
popara96: Agreed, I modified your suggestion a bit and split it into more rows for IDE readability.

JoC0de (approved):
👍

PR #577, opened 2023-10-06
Shortening the path of source files in source code nuget packages

popara96:
This PR will sort out long paths made by source code packages by unpacking all source files inside the packageDirectory/Sources.

### DIFF
--- a/src/NuGetForUnity.Tests/Assets/Tests/Editor/NuGetTests.cs
+++ b/src/NuGetForUnity.Tests/Assets/Tests/Editor/NuGetTests.cs
@@ -775,6 +775,37 @@ public class NuGetTests
             package.Version);
     }
 
+    [Test]
+    [TestCase("Microsoft.Azure.WebJobs.Sources", "3.0.37")]
+    [TestCase("NServiceBus.Testing.Fakes.Sources", "7.1.13")]
+    public void TestSourceCodePackageInstall(string packageId, string packageVersion)
+    {
+        var package = new NugetPackageIdentifier(packageId, packageVersion) { IsManuallyInstalled = true };
+        Assume.That(InstalledPackagesManager.IsInstalled(package), Is.False, "The package IS installed: {0} {1}", package.Id, package.Version);
+
+        NugetPackageInstaller.InstallIdentifier(package);
+        Assert.IsTrue(
+            InstalledPackagesManager.IsInstalled(package),
+            "The package was NOT installed: {0} {1}",
+            package.Id,
+            package.Version);
+
+        var packageDirectory = Path.Combine(
+            ConfigurationManager.NugetConfigFile.RepositoryPath,
+            $"{package.Id}.{package.Version}");
+
+        Assert.That(Path.Combine(packageDirectory, "Sources"), Does.Exist);
+        Assert.That(Path.Combine(packageDirectory, "contentFiles"), Does.Not.Exist);
+
+        // cleanup and uninstall everything
+        NugetPackageUninstaller.UninstallAll(InstalledPackagesManager.InstalledPackages.ToList());
+        Assert.IsFalse(
+            InstalledPackagesManager.IsInstalled(package),
+            "The package is STILL installed: {0} {1}",
+            package.Id,
+            package.Version);
+    }
+
     private static void ConfigureNugetConfig(InstallMode installMode)
     {
         var nugetConfigFile = ConfigurationManager.NugetConfigFile;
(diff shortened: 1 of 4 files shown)


src/NuGetForUnity/Editor/NugetPackageInstaller.cs
```diff
@@ -196,36 +198,125 @@ private static bool Install([NotNull] INugetPackage package, bool refreshAssets,
...
+                                    "Selecting target framework directory 'any' and language 'any' as best match for the package {1}", package);
+
+                                PackageContentManager.ExtractPackageSources(anyFrameworks, baseDirectory);
+                            }
+                            else
+                            {
+                                Debug.LogWarningFormat(
+                                    "Couldn't find a source code folder with a supported target-framework for the package {0}",
+                                    package);
+                            }
```
igor84, line 319: You can simplify this using something like:
```cs
if (csFiles.Count > 0)
{
	TryExtractBestFrameworkSources(csFiles, 'cs');
}
else if (anyFiles.Count > 0)
{
	TryExtractBestFrameworkSources(anyFiles, 'any');
}

private void TryExtractBestFrameworkSources([NotNull] [ItemNotNull] IReadOnlyCollection<KeyValuePair<string, List<ZipArchiveEntry>> frameworks, [NotNull] string sourceDirName)
{
	var bestFrameworkMatch = TargetFrameworkResolver.TryGetBestTargetFramework(frameworks, framework => framework.Key);
	var frameworkKey = bestFrameworkMatch.Key ?? "any";
	if (frameworks.TryGetValue(frameworkKey, out var bestFramework))
	{
		NugetLogger.LogVerbose(
			"Selecting target framework directory '{0}' and language '{1}' as best match for the package {2}",
			bestFramework,
			sourceDirName,
			package);

		PackageContentManager.ExtractPackageSources(bestFramework, baseDirectory);
	}
	else
	{
		Debug.LogWarningFormat(
			"Couldn't find a source code folder with a supported target-framework for the package {0}",
			package);
	}
}
```

src/NuGetForUnity/Editor/PackageContentManager.cs
```diff
@@ -194,12 +203,47 @@ internal static bool ShouldSkipUnpackingOnPath([NotNull] string path)
...
+                {
+                    if (entryFullName[j] != lastCommonDir[j])
+                    {
+                        lastCommonDir = entryFullName.Substring(0, j);
+                        break;
+                    }
+                }
+            }
+
+            var sourcesDirectory = Path.Combine(baseDir, "Sources").Replace('\\', '/');
```
igor84, line 228: Why do you need this Replace here? I don't see you use this variable later in a way that requiers it.

igor84 (approved):

src/NuGetForUnity/Editor/PackageContentManager.cs
popara96: It was useful when I was counting `/` occurrences, I will remove it now

JoC0de (changes requested):
It would be nice to at least have one NuGet package that has source code to be installed by the Test so we can ensure this woks / doesn't brake later.
We could even create a ZipFile on-the-fly just for testing instead of using a existing .nupgk

src/NuGetForUnity/Editor/NugetPackageInstaller.cs
JoC0de: yes without the "extraction of a function" it is definitely to much copy-paste.

popara96: Simplified but without `[ItemNotNull]` as Rider said `Applying nullness annotation to value type is meaningless`.

popara96:
Will write tests tomorrow.

src/NuGetForUnity.Tests/Assets/Tests/Editor/NuGetTests.cs
```diff
@@ -775,6 +775,37 @@ public void TestSlimRestoreInstall(string packageId, string packageVersion)
             package.Version);
     }
 
+    [Test]
+    [TestCase("Microsoft.Azure.WebJobs.Sources", "3.0.37")]
+    [TestCase("NServiceBus.Testing.Fakes.Sources", "7.1.13")]
+    public void TestSourceCodePackageInstall(string packageId, string packageVersion)
```
popara96, line 781: I added a check to see if `Sources` directory is added in the package root and ensured that `contentFiles` directory doesn't exist. I'm trying to thing of other generic checks that we could do while installing this but nothing comes to mind as directories/files under `Sources` will be very package-specific. H

JoC0de (approved):
I didn't see any more issues, I would only fix the formatting at some places. See e.g. the Compiler warnings about trailing whitespace.

PR #582, opened 2023-10-13
Various fixes and improvements

popara96:
I apologize upfront for a bit larger PR. Wanted to fix a couple of small things but one thing led to another and we ended up with this. Basically the only 'larger' change here is how the `Update` tab works, and how the tab responsibilities are split, other than that it's really just minor fixes.

Changes introduced:

1. ### **Fixed `refreshAssets` propagation**
    We are now refreshing assets only for the topmost package when installing/updating packages. This will make sure that while installing or updating we refresh the assets only once. It also fixes one dependency being marked as explicit wrongly during installation.

2. ### **Fixed update tab for V2 packages**
    Until now every version of the package newer than the version installed was shown as a separate package. This is now fixed by showing the dropdown selection list (similar to how V3 looks), with stacking `Release Notes` in the `Details` section showing release notes for every newer version in descending order.

3. ### **Cleaner tab responsibilities**
    Every tab now has a separate reponsibility depending on what it should do: 

    - `Online` tab will show dropdown list of versions along with `Install` button if the package is not installed, otherwise it will only show `Installed Version` to indicate what is already installed.
    
    - `Installed` tab shows what version is installed and options to uninstall packages or mark dependencies as explicit packages.
    
    - `Updates` tab shows installed version and shows, in a dropdown list, selection of newer versions with `Update` button (if `Show Downgrades` is not checked), or lower versions with `Downgrade` button (if `Show Downgrades` is checked).
    
    - `Update All` button is now shown only if `Show Downgrades` is not checked as we don't want to offer downgrading all packages at once. Default selected versions in dropdown lists are always the highest newer or highest lower version, depending on the checkbox, and the dropdown does NOT contain the currently installed version.

    **_NOTE:_** when fetching packages, list of versions is now packed in _descending_ instead of _ascending_ order. This is done so we don't have to sort them by descending order every time we refresh the dropdown list of available packages, so as not to throttle the UI.

4. ### **Fixed finding best framework group dependencies**
    Small fix that caches the framework group after we first find it, this also fixes logging the best target framework group every time we expand `Details` on a package.


src/NuGetForUnity/Editor/Ui/NugetWindow.cs
```diff
@@ -15,6 +15,15 @@
 
 namespace NugetForUnity.Ui
 {
+    public enum NugetWindowTab
```
JoC0de, line 18: Please don't declare multiple public types / enums in one file. Either move it to a separate file or use a nested private type / enum.

src/NuGetForUnity/Editor/Ui/NugetWindow.cs
```diff
@@ -849,14 +872,24 @@ private void DrawPackage(INugetPackage package, GUIStyle packageStyle, GUIStyle
             var installedPackages = InstalledPackagesManager.InstalledPackages;
             var installed = installedPackages.FirstOrDefault(p => p.Id == package.Id);
 
-            EditorGUILayout.BeginHorizontal();
+            // if we are on the update tab, we do not want to show packages that have no updates if we're showing updates; similarly, we do not
+            // show packages that are on the lowest possible version if we're showing downgrades
+            if (currentTab == NugetWindowTab.UpdatesTab &&
+                installed != null &&
```
JoC0de, line 878:
```suggestion
                installed != null &&
                package.Versions.Count >= 1 &&
```
It shouldn't be possible because the list should at least contain the package version itself, but I would still check it so this doesn't throw a exception.

src/NuGetForUnity/Editor/Ui/NugetWindow.cs
```diff
@@ -643,9 +658,10 @@ private void DrawPackages(IEnumerable<INugetPackage> packages, bool canBeSelecte
 
             foreach (var package in packages)
             {
-                EditorGUILayout.BeginVertical(backgroundStyle);
-                DrawPackage(package, backgroundStyle, contrastStyle, canBeSelected);
-                EditorGUILayout.EndVertical();
+                using (new EditorGUILayout.VerticalScope(backgroundStyle))
```
JoC0de, line 661: Is it really supported in Unity 2018? Because the doc doesn't list it:
image
But the CI should have failed if it is not supported.

src/NuGetForUnity/Editor/Ui/NugetWindow.cs
```diff
@@ -934,44 +967,71 @@ private void DrawPackage(INugetPackage package, GUIStyle packageStyle, GUIStyle
...
-                if (package.Versions.Count <= 1)
+                // Show the installed version of the package if it exists
+                if (installed != null)
                 {
-                    GUILayout.Label($"Version {package.PackageVersion.FullVersion}");
+                    GUILayout.Label($"Installed Version: {installed.PackageVersion.FullVersion}");
                 }
-                else
+
+                // Show the version selection dropdown only on Updates tab OR on Online tab if the package is not installed and not already in Unity
```
JoC0de, line 977: Shouldn't it be possible to update a package from the "Online tab" or even from the "Installed tab"?
E.g. in Visual Studio you can update / uninstall package from all 3 tabs: 
Visual Studio Installed tab:
image
Visual Studio Online tab:
image
Visual Studio Update tab:
image

src/NuGetForUnity/Editor/Ui/NugetWindow.cs
```diff
@@ -934,44 +967,71 @@ private void DrawPackage(INugetPackage package, GUIStyle packageStyle, GUIStyle
...
+                                out _,
+                                out var maxWidth);
+
+                            if (installed != null)
+                            {
+                                // we're showing only lower or only higher versions, depending on the 'Show Downgrades' checkbox.
+                                // This can only happen in Updates tab.
+                                if (showDowngrades)
+                                {
+                                    package.Versions.RemoveAll(version => version >= installed.PackageVersion);
```
JoC0de, line 1003: I don't think it is a good idea to manipulate the information of the package directly I would create a new list containing only the package version's that should be visible in the UI. Imagine a package that depends on a version you removed from the "UI" so it can't be installed as you removed the information that the version exists.
I think this should also remove the need to re-download all packages by calling `UpdateUpdatePackages` when the user toggles the `showDowngrades`.

src/NuGetForUnity/Editor/Ui/NugetWindow.cs
```diff
@@ -934,44 +967,71 @@ private void DrawPackage(INugetPackage package, GUIStyle packageStyle, GUIStyle
...
+                                    package.Versions.RemoveAll(version => version <= installed.PackageVersion);
+                                }
+                            }
+
+                            versionDropdownData = new VersionDropdownData { SortedVersions = package.Versions, CalculatedMaxWith = maxWidth + 5, };
+
+                            versionDropdownData.DropdownOptions = versionDropdownData.SortedVersions.Select(version => version.FullVersion).ToArray();
+
+                            // Show the highest available update/downgrade first
+                            versionDropdownData.SelectedIndex = 0;
```
JoC0de, line 1016: On the Update / Online tab I would show the highest version, on the installed we can show the current selected. See Visual Studio package window.

src/NuGetForUnity/Editor/Ui/NugetWindow.cs
```diff
@@ -980,29 +1040,32 @@ private void DrawPackage(INugetPackage package, GUIStyle packageStyle, GUIStyle
...
+                            {
+                                NugetPackageUpdater.Update(installed, package);
+                                UpdateInstalledPackages();
+                                UpdateUpdatePackages();
+                            }
                         }
                     }
 
-                    if (GUILayout.Button("Uninstall"))
+                    if (currentTab == NugetWindowTab.InstalledTab && GUILayout.Button("Uninstall"))
```
JoC0de, line 1068: Se above I would allow to uninstall package even on Update / Online tab. See Visual Studio package window. Or is the UI to bloated?

src/NuGetForUnity/Editor/Ui/NugetWindow.cs
```diff
@@ -980,29 +1040,32 @@ private void DrawPackage(INugetPackage package, GUIStyle packageStyle, GUIStyle
                         }
                     }
 
-                    var versionComparison = installed.PackageVersion.CompareTo(package.PackageVersion);
-                    if (versionComparison < 0)
+                    if (currentTab == NugetWindowTab.UpdatesTab)
```
JoC0de, line 1043: Se above I would allow to update package even on Installed / Online tab. See Visual Studio package window. Or is the UI to bloated?

JoC0de (changes requested):
The PR actually isn't that big I had bigger ones 😄
It was not that much of a problem to see what change was related to which feature.
I like the caching of the selected dependencies. Eventually we can even extend this to provide warnings if a package has no supported dependency group / e.g. a user tries to install a .Net Framework package into a .Net Standard project.
For the UI changes I am not sure if we really should remove the possibility to update / uninstall packages on the other tabs. I added some screenshots of how it works in the Visual Studio package manager window.
What is the reasoning behind adding the `showDowngrades` toggle? If I have a dropdown I know what version to select / or are there to many versions and the Unity dropdown has a to bad usability?
If we Update the UI we should also update the screenshots used in the `Readme.md`.

src/NuGetForUnity/Editor/Ui/NugetWindow.cs
popara96: I discussed this with @igor84 and he would need it to be public for the upcoming plugin expansion feature. So from these two options, I would probably move it to a separate file. Leaving this conversation unresolved, waiting for your response.

src/NuGetForUnity/Editor/Ui/NugetWindow.cs
popara96: It does work, I am testing all of the code in Unity 2018.4.36f1 :)

src/NuGetForUnity/Editor/Ui/NugetWindow.cs
popara96: The idea behind this was to reduce the number of buttons, as the UI as it currently is gets a bit too squashed already (I can see that VS's Nuget has a separate menu for these types of package manipulation), and at the same time it seems more intuitive to look to `Installed` tab if you want to uninstall, `Online` tab if you're looking for a new package and `Updates` tab if you're looking to get some other version of an already installed package.
 I will update the screenshots accordingly when we settle on the design.

src/NuGetForUnity/Editor/Ui/NugetWindow.cs
popara96: I will fix this by simply filtering what goes into `DropdownOptions` better, and on toggling I will just clear `versionDropdownDataPerPackage`.

src/NuGetForUnity/Editor/Ui/NugetWindow.cs
igor84: You can move it to a separate file.

src/NuGetForUnity/Editor/Ui/NugetWindow.cs
igor84: We noticed buttons with label about the version next to them often start to overlap with package ID from the left. 

Also we decided to add Show downgrades check box for a couple of reasons. First we assumed that we want to expose the downgrade functionality and definitely have it on updates tab. Before this commit only newer versions were fetched for both V2 and V3 API and we changed V3 to includeAllVersions. Since V2 has no such easy API we left it working as always.

When we added that we realized that now updates tab almost always shows all the installed packages some actually having new versions but all of them having some old versions and most often we go to that tab to easily see what packages actually have newer versions. Also we often actually want to use update all button to update all the packages that have new version. If we didn't hide downgrade options it might look like that button would update some packages but downgrade others since it would actually update to the versions you selected next to each package.

This seemed as a confusing UX so we hoped to make it more streamlined for the most often needed actions while still giving user options to do advanced stuff.

JoC0de: Ok I understand. In Visual Studio the Updates tab can also only update. They use the Installed tab to downgrade. And with the amount of text / Bottons inside a single row I also noticed this so I agree with the splitting.

JoC0de (changes requested):

JoC0de:
Nice 👍  with the screenshots. Only some portions of te readme.md need to be updated to conform with the new UI and then we can merge it.

JoC0de (approved):

JoC0de:
Nice 👍 you even changed more in the Readme than I expected 👍 
I fixed the merge conflict + run autoformatter script so we can merge (from my side)

popara96:
@JoC0de We had a bug where we could select different versions of a package because of differently calculated hash code, so we could have the list filled with the same package but different versions and we don't want that. It's split now and this way we also have selected packages remembered across 'Installed' tab and 'Updates' tab with updates/downgrades.

JoC0de (approved):
I fixed a bug: you didn't used `selectedPackageUninstalls` so uninstall selected didn't worked.
I also added switched to HashSet with custom comparer I think it is much easier to handle.
I didn't tested it 😃 , so could you pleas test it.

PR #591, opened 2023-10-21
allow overwriting nuget cache location

### LINKED ISSUE #588: NuGetForUnity.Cli crashes if `Environment.SpecialFolder.LocalApplicationData` is not set.
## Description
I have been running this tool inside of a GitHub action and it has worked great, but we recently switched to using Kubernetes runners and suddenly this action started crashing with the error:
```
Unhandled exception. System.TypeInitializationException: The type initializer for 'NugetForUnity.NugetHelper' threw an exception.
 ---> System.ArgumentException: The value cannot be an empty string. (Parameter 'path')
   at System.ArgumentException.ThrowNullOrEmptyException(String argument, String paramName)
   at System.IO.Path.GetFullPath(String path)
   at NugetForUnity.NugetHelper..cctor() in /home/runner/work/NuGetForUnity/NuGetForUnity/src/NuGetForUnity/Editor/NugetHelper.cs:line 56
   --- End of inner exception stack trace ---
   at NugetForUnity.NugetHelper.set_InstallDependencies(Boolean value) in /home/runner/work/NuGetForUnity/NuGetForUnity/src/NuGetForUnity/Editor/NugetHelper.cs:line 111
   at NuGetForUnity.Cli.Program.Main(String[] args) in /home/runner/work/NuGetForUnity/NuGetForUnity/src/NuGetForUnity.Cli/Program.cs:line 55
```

The cause from the issue seems to be [this line](https://github.com/GlitchEnzo/NuGetForUnity/blob/v3.1.3/src/NuGetForUnity/Editor/NugetHelper.cs#L56C10-L56C10) that is looking to get the `Path.GetFullPath` of `Environment.SpecialFolder.LocalApplicationData`. The error is caused when the `LocalApplicationData` folder is null or empty. This can occur in some situations where the tool is being run by the `root` user on Linux.

It seems like the issue will still occur in the latest version in [PackageCacheManager.cs](https://github.com/GlitchEnzo/NuGetForUnity/blob/9b903a977223bfce722272901b4f579d5365b33d/src/NuGetForUnity/Editor/PackageCacheManager.cs#L18C13-L18C13).

A workaround was to export the environment variable `XDG_DATA_HOME` as documented [here](https://developers.redhat.com/blog/2018/11/07/dotnet-special-folder-api-linux), and make sure that the folder exists.

However, there still should either be a fallback cache path or this exception should be properly handled so that it is clear what the issue is.

-   **NuGetForUnity.Cli Version:** 3.1.3
-   **Unity Version:** N/A
-   **Operating System:** Linux Container `mcr.microsoft.com/dotnet/sdk:7.0`

JoC0de:
fixes #588
* safely handle if 'localappdata' folder is not specified
* allow overwriting nuget cache location

### DIFF
--- a/src/NuGetForUnity/Editor/PackageCacheManager.cs
+++ b/src/NuGetForUnity/Editor/PackageCacheManager.cs
@@ -1,9 +1,21 @@
-﻿using System;
+﻿#pragma warning disable SA1512,SA1124 // Single-line comments should not be followed by blank line
+
+using System;
 using System.IO;
 using JetBrains.Annotations;
 using NugetForUnity.Configuration;
 using NugetForUnity.Models;
 using NugetForUnity.PackageSource;
+
+#region No ReShaper
+
+// ReSharper disable All
+// needed because 'JetBrains.Annotations.NotNull' and 'System.Diagnostics.CodeAnalysis.NotNull' collide if this file is compiled with a never version of Unity / C#
+using SuppressMessageAttribute = System.Diagnostics.CodeAnalysis.SuppressMessageAttribute;
+
+// ReSharper restore All
+
+#endregion
 
 namespace NugetForUnity
 {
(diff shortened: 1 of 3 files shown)


src/NuGetForUnity/Editor/PackageCacheManager.cs
```diff
@@ -156,5 +153,24 @@ private static INugetPackage GetInstalledPackage([NotNull] INugetPackageIdentifi
                 installedPackage.Version);
             return installedPackage;
         }
+
+        private static string DetermineCacheOutputDirectory()
+        {
+            const string cacheEnvironmentVariableName = "NuGetCachePath";
+            var cacheFromEnvironment = Environment.GetEnvironmentVariable(cacheEnvironmentVariableName);
+            if (!string.IsNullOrEmpty(cacheFromEnvironment))
```
igor84, line 161: In case the path given in env doesn't exist should we try to CreateDirectory or just fallback to default handling bellow? Although we might want to add some handling if default path doesn't exist as well.

igor84 (approved):

src/NuGetForUnity/Editor/PackageCacheManager.cs
JoC0de: `Directory.CreateDirectory(CacheOutputDirectory);` is called inside the static constructor for both cases.
I think this is the expected behavior.
Poth path's / the name of the environment variable are the once used by the official NuGet client prior to 3.5. Eventually we can switch to the newer cache format / location later: [docu](https://learn.microsoft.com/en-us/nuget/consume-packages/managing-the-global-packages-and-cache-folders)

PR #593, opened 2023-10-21
update version to 4.0.0

JoC0de:


JoC0de:
@igor84 @popara96 
I think we should dump the version to `4.0.0` instead of only `3.2.0` because we have so many even brackig changes. Or what do you think?

igor84:
I agree. I think we have two more important things we want to do: selecting multiple packages for installation on Online tab and the plugin support. Plugin support should be done by Wednesday but either way those can be released in some next minor 4.0.1 release.

igor84 (approved):

Issue #564, opened 2023-09-11
Add support for NugetForUnity plugins


igor84:
## Why would this be useful

Sometimes users would like to be able to add custom functionality to NugetForUnity that is specific to their workflow, or packages they use and they don't see that it would making sense to add it for everyone in NugetForUnity itself.

For example in my company we implemented a "Link source" button next to each package that knows how we keep our projects organized and can automatically find the proper source folder of the given package and replace its folder with a link (windows directory junction) to that folder so that you can edit the source directly from your project.

Or you want to support installing certain packages with native libraries in a some specific way so you can add custom code that will just work with packages you need. We could even provide our own plugin that handles a few most popular nuget packages with native libraries. This could help with issue #419.

## How it could be implemented

We would need a separate class library project that would only contain interfaces that we need in order for plugin to communicate with the main NugetForUnity code. That project would need to be built into a dll and installed as a dll with NugetForUnity. It has to be a dll because the plugins need to be dlls so that they can also be loaded from the CLI version.

NugetForUnity could have a setting telling where to look for plugins in the current project. It might even be useful to list and explicitly enable plugins in the preferences window so you can have an option of having plugin code in the project but temporarily deactivating it.

The CLI version would manually load the dlls from that folder and in Unity we would just get a reference to Assembly objects that Unity already loaded and in them we could find a class that has some `[NugetPlugin]` attribute or implements some INugetPlugin interface for example. It would use some kind of Register calls so that each plugin can register callbacks it is interested in so that it can add additional buttons to DrawPackage method, or handle how are some files extracted or cleaned up in its own way.

We would then just need to define what extension points we want to provide and implement those, plus we should make it so that new extension points can be added in the future without breaking backwards compatibility.

JoC0de:
I just realized that we need to version the `NugetForUnity.PluginAPI` separately from the normal `NugetForUnity` as if we upload the `NugetForUnity.PluginAPI` as a NuGet package it would contain a version number different than the version number of the `NugetForUnity.PluginAPI.dll` that we have in the GIT repository / UPM package. So the user will get version conflict warnings. If I have some time I will create a PR. But until this is fixed I wouldn't create a official release.

igor84:
I'm not sure I understand. Is this related to that option in Unity that should be disabled so it doesn't check versions between DLLs? I think we will have to require this because we can't expect third party plugins to all be updated when we update the Plugin API.

JoC0de:
[...]

PR #596, opened 2023-10-24
Added support for PluginAPI #564

igor84:


.github/workflows/main.yml
```diff
@@ -65,11 +65,14 @@ jobs:
         with:
           version: ${{ needs.determineVersionNumber.outputs.version }}
 
-      - name: Upload NuGetForUnity.dll
+      - name: Upload NuGetForUnity dlls
         uses: actions/upload-artifact@v3
         with:
           name: NuGetForUnity.dll
-          path: ./src/NuGetForUnity.CreateDll/bin/Release/NugetForUnity.dll
+          path: |
```
igor84, line 72: @JoC0de Can you help me figure out why the build fails? This seems like the only change that might cause the builds to fail as they do but I don't understand why. I wrote this according to documentation for upload-artifactory action. Or maybe I am missing something in the build logs and the issue is elsewhere?

igor84: Ok, I figured out you added a check that only allows you to change these files, but I added myself there now as well and it is still failing?

JoC0de: @igor84 The GitHub Action is triggered by a `pull_request_target` event so it works on the code base of the target of the PR in this case the `master` branch. A normal `pull_request` can't be used to checkout forks as it would allow everyone to change the code that runs on the CI. I explicitly checkout the code from the merged PR but protect the CI files from being modified to mitigate the risk of anyone being able to change it.
With the current CI only I can edit `action` files as the are also fetched from the merged PR.
The `.github/workflows/main.yml` file currently is always fetched from the `master` branch so changes are not reflected until merged.
I think the only way to run the GitHub Action with your changes is if you rename your branch to match the pattern (`feature/*`):
https://github.com/GlitchEnzo/NuGetForUnity/blob/91dcd516c9b441d7c68c07158c9f12be13edbaea/.github/workflows/main.yml#L3-L6
Then the CI does a single checkout of the code from the state of the `feature/*` branch and directly executes it.

igor84: I created a new `feature/plugin-support` branch and made a pull request from it but it still fails?

igor84: Maybe the easiest solution would be to commit the change with my username in action.yml directly to master?

JoC0de: No it works. Now every action is triggered twice:
image
And the `push` is from your branch. And the are working....

igor84: Ah, ok. Then I will delete this pull request and its branch. Thank you.

JoC0de: Even if I commit it I can't change content of the `main.yml` and get it "executet" by a `pull_request_target` event.

PR #597, opened 2023-10-24
Added support for PluginAPI #564

igor84:


JoC0de:
I don't like binaries inside GIT, they can get out-of-sync.
Shouldn't it be possible without `.dll` inside GIT?
- Place the PluginAPI source code inside a subdirectory of `src/NuGetForUnity/Editor/` so Unity knows it and it is included inside the UPM package
- The `NuGetForUnity.Cli` and `NuGetForUnity.CreateDll` shouldn't have a issue with it
- Create a `NuGetForUnity.PluginAPI.csproj` that we can build separately and provide as a download / publish as a NuGet package

So others can create the Plugin either by:
- installing the NuGet package
- copying the DLL (not recommended) 
- directly inside Unity

igor84:
Bellow is my reasoning for having to include dll. Correct me if I understood something wrong:

We encourage installing NugetForUnity as a UPM package. Installing that way would install the contents of Editor folder as source into user's project. That means PluginAPI would be installed as source as well. But I created some plugin that is compiled into a dll and inside that dll it is recorded that it depends on NugetForUnity.PluginAPI.dll but now there is no such dll in the project.

Maybe this would work if we can make Unity compile asmdef for PluginAPI before it tries to load the plugin dll but I am not sure that is possible.

JoC0de:
So the goal is to be able to implement the plugin in a external .csproj and not being limited to only implement it inside the same or a different unity project?

igor84:
The plugin has to be a dll (compiled from external csproj) so that it can also be loaded from CLI version.

.github/workflows/main.yml
igor84, line 1: Should I just copy the packageTool task from this file and create a packagePluginApi task from it just targeting a different csproj in order to publish it as a nuget package?

.github/workflows/main.yml
```diff
@@ -65,15 +65,24 @@ jobs:
...
+            ./src/NuGetForUnity.CreateDll/bin/Release/NuGetForUnity.dll
+            ./src/NuGetForUnity.CreateDll/bin/Release/NuGetForUnity.PluginAPI.dll
+            ./src/NuGetForUnity.CreateDll/bin/Release/NuGetForUnity.PluginAPI.xml
           if-no-files-found: error
 
+      - name: Ensure NuGetForUnity.PluginAPI.dll in GIT is up-to-date (if it fails pleas build the NugetForUnity.PluginAPI.csproj and upload the changed files to git)
+        run: git diff --exit-code ./src/NuGetForUnity/Editor/PluginAPI/NuGetForUnity.PluginAPI.dll
+
+      - name: Ensure NuGetForUnity.PluginAPI.xml in GIT is up-to-date (if it fails pleas build the NugetForUnity.PluginAPI.csproj and upload the changed files to git)
+        run: git diff --exit-code ./src/NuGetForUnity/Editor/PluginAPI/NuGetForUnity.PluginAPI.xml
```
igor84, line 82: What is the intent here? Will it check that each commit has changes in these files? They should only be updated if something in PluginAPI project was changed which should be pretty rare.

JoC0de: No they should only check if `NuGetForUnity.PluginAPI.dll` is updated if something it the code of the PluginAPI project was changed. It currenlty has the issue that the `.dll` contains the version number so it always failes.

igor84:
@JoC0de thank you very much for helping polish this PR :muscle:

JoC0de (approved):
So I am done with my changes 😄 
The implementation looks good 👍 
I already updated the version number in the source code to `4.0.2` so we can release it after merging.

PR #606, opened 2023-11-14
Updated plugin-dev-readme

igor84:
I updated the plugin-dev-readme to include instructions to use NuGetForUnity.PluginAPI nuget package when developing plugins.

### DIFF
--- a/README.md
+++ b/README.md
@@ -270,7 +270,7 @@ Restore nuget packages of a single Unity Project: `dotnet nugetforunity restore 
 
 NugetForUnity has plugin support. If you open the NugetForUnity section in Unity preferences it will list the plugins you have installed in your project and you can enable them from there.
 
-Plugins are any dlls which contain NugetForUnityPlugin in their name and have a class inside them that implements the `INugetPlugin` interface. They can be placed anywhere inside the project (inside the `Assets` directory) and can even be installed as a nuget package.
+Plugins are any dlls which contain NugetForUnityPlugin in their name and have a class inside them that implements the `INugetPlugin` interface. They can be placed anywhere inside the project (inside the `Assets` directory).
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -270,7 +270,7 @@ Restore nuget packages of a single Unity Project: `dotnet nugetforunity restore
 
 NugetForUnity has plugin support. If you open the NugetForUnity section in Unity preferences it will list the plugins you have installed in your project and you can enable them from there.
 
-Plugins are any dlls which contain NugetForUnityPlugin in their name and have a class inside them that implements the `INugetPlugin` interface. They can be placed anywhere inside the project (inside the `Assets` directory) and can even be installed as a nuget package.
+Plugins are any dlls which contain NugetForUnityPlugin in their name and have a class inside them that implements the `INugetPlugin` interface. They can be placed anywhere inside the project (inside the `Assets` directory).
```
JoC0de, line 273:
```suggestion
Plugins are any dlls which contain NugetForUnityPlugin in their name and have a class inside them that implements the `INugetPlugin` interface. They can be placed anywhere inside the project, inside the `Assets` directory, therefore they can also be installed as a NuGet package by themselves.
```
Not 100% sure by I think the information is helpful

JoC0de (approved):

README.md
igor84: I removed the part about plugin being installed as a nuget package because I realized it would cause issues as it is. For example if you are gitignoring the folder where you install packages because you are counting on NugetForUnity to download them for you then during CI build the packages would actually be installed wihout the plugin enabled because at the start up time there wouldn't yet be a plugin dll in the project.

JoC0de: We could probably fix this by changing the order in witch the NuGet packages are restored and than force a Plugin reload. But for now we can merge the current changes.

PR #611, opened 2023-12-25
Implemented support for selecting multiple packages for installation

igor84:
image

image


popara96 (approved):

src/NuGetForUnity/Editor/Ui/NugetWindow.cs
```diff
@@ -699,6 +769,48 @@ private void DrawOnline()
             EditorGUILayout.EndScrollView();
         }
 
+        private void DrawSelectedForInstallationHeader(GUIStyle headerStyle)
+        {
+            var rectangle = GUILayoutUtility.GetRect(GUIContent.none, headerStyle, GUILayout.Height(25f));
+
+            EditorGUI.LabelField(rectangle, string.Empty, headerStyle);
+
+            var arrow = showPackagesToInstall ? "\u2227" : "\u2228";
```
JoC0de, line 778: I would have used some constants for the string so we don't have "magic numbers"

JoC0de (approved):
sorry for the delay
looks good to me 👍
only one small note

igor84:
I realized I didn't properly handle packages that are already installed or are available in Unity in the selection code so I fixed that. I also added a "Select all from clipboard" button so users can copy a list of package ids they need to install from somewhere and just click this button to have them all selected for installation.

Issue #615, opened 2024-01-18
Unintended behavior when there are analyzers for multiple Roslyn versions in one nuget package


hadashiA:
## Description

A single nuget package can contain multiple Roslyn analyzers with same content but different Roslyn versions.
This is to support older versions of Roslyn  users.

Currently, this seems to be a problem in some cases.

For example, if we install `System.Text.Json` with NuGetForUnity,   source generators x3 with the same content are imported.
 
![スクリーンショット 2024-01-18 16 54 03](https://github.com/GlitchEnzo/NuGetForUnity/assets/727159/dfde2993-5898-42c8-bf2c-4ada4387cb1c)

```
System.Text.Json 8.0.1/
└── analyzers/
    └── dotnet/
        ├── roslyn3.11/
        │   └── cs/
        │       └── System.Text.Json.SourceGeneration.dll < For Roslyn 3.11 or newer
        ├── roslyn4.0/
        │   └── cs/
        │       └── System.Text.Json.SourceGeneration.dll < For Roslyn 4.0 or newer
        └── roslyn4.4/
            └── cs/
                └── System.Text.Json.SourceGeneration.dll  < For Roslyn 4.4 or newer
```

I tried the following simple System.Text.Json Source Generator example,  getting errors.

```cs
public class WeatherForecast
{
    public DateTime Date { get; set; }
    public int TemperatureCelsius { get; set; }
    public string? Summary { get; set; }
}

[JsonSourceGenerationOptions(WriteIndented = true)]
[JsonSerializable(typeof(WeatherForecast))]
internal partial class SourceGenerationContext : JsonSerializerContext
{
}
```

![スクリーンショット 2024-01-18 14 48 24](https://github.com/GlitchEnzo/NuGetForUnity/assets/727159/c0a616b1-191f-41d7-a9d0-cb28cde06b2d)

This is because multiple SourceGenerators with the same contents are run.

[Describe the issue you have. If applicable provide logs produced by enabeling verbose logging in the NuGetForUnity settings.]

-   **NuGet Package:** System.Text.Json
-   **NuGetForUnity Version:** 4.0.2
-   **Unity Version:** 2022.2.21f, etc
-   **Operating System:** macOS

##  Improvement proposal

This problem can be solved by limiting the number of dlls with the `RoslynAnalyzer` asset label to one.
Currently, analzyer is disable on any platform, so doing this was sufficient.

[...]

PR #616, opened 2024-01-18
Support for nuget package with multiple Roslyn version analyzers

hadashiA:
To address the #615, I added a process to prevent the same analyzer from being enabled more than once.

It works as follows.

- If `analyzers/dotnet` subfolders are not split into versions, then it remains the same.
- If not,  only newer versions that can be supported will be given the `RoslynAnalyzer` label.


src/NuGetForUnity/Editor/NugetAssetPostprocessor.cs
```diff
@@ -191,6 +221,44 @@ private static void ModifyImportSettingsOfRoslynAnalyzer([NotNull] PluginImporte
...
             AssetDatabase.SetLabels(plugin, new[] { RoslynAnalyzerLabel, ProcessedLabel });
 
+            // The nuget package can contain analyzers for multiple Roslyn versions.
+            // In that case, for the same package, the most recent version must be chosen out of those available for the current Unity version.
+            var assetRoslynVersion = GetRoslynVersionNumberFromAnalyzerPath(plugin.assetPath);
+            if (assetRoslynVersion != null)
+            {
+                var versionPrefixIndex = plugin.assetPath.IndexOf(AnalyzersRoslynVersionsFolderName, StringComparison.Ordinal);
+                var analyzersVersionsRoot = JoinPathComponents(plugin.assetPath.Substring(0, versionPrefixIndex), AnalyzersRoslynVersionsFolderName);
+                var analyzersFolders = AssetDatabase.GetSubFolders(analyzersVersionsRoot);
```
igor84, line 231: Can we be sure `GetSubFolders` will work at this point since this is called from `OnPreprocessAsset` meaning maybe not all sub folders are yet imported?

src/NuGetForUnity/Editor/NugetAssetPostprocessor.cs
```diff
@@ -180,7 +196,21 @@ private static string GetNuGetRepositoryPath()
...
+        private static string GetRoslynVersionNumberFromAnalyzerPath(string analyzerAssetPath)
+        {
+            var versionPrefixIndex = analyzerAssetPath.IndexOf(AnalyzersRoslynVersionSubFolderPrefix, StringComparison.Ordinal);
+            if (versionPrefixIndex < 0)
+            {
+                return null;
+            }
+
+            return analyzerAssetPath.Substring(versionPrefixIndex + AnalyzersRoslynVersionSubFolderPrefix.Length)
+                .Split(Path.DirectorySeparatorChar)
```
igor84, line 209: Doesn't unity always reports asset paths with forward slash no matter what operating system you are on. In other words I think this would work on MacOS and Linux but not on Windows since you would try to split on `\` while the asset path will use `/`. Note that in the other places where we used DirectorySeparatorChar we only did that after putting the path through `Path.GetFullPath()` method which normalized the separators to current OS.

src/NuGetForUnity/Editor/NugetAssetPostprocessor.cs
```diff
@@ -191,6 +221,44 @@ private static void ModifyImportSettingsOfRoslynAnalyzer([NotNull] PluginImporte
...
+                        return false;
+#endif
+                    })
+                    .ToArray();
+
+                // If most recent valid analyzers exist elsewhere, remove label `RoslynAnalyzer`
+                if (!enabledRoslynVersions.Contains(assetRoslynVersion) ||
+                    string.CompareOrdinal(assetRoslynVersion, enabledRoslynVersions.Max()) < 0)
+                {
+                    AssetDatabase.SetLabels(plugin, new[] { ProcessedLabel });
```
igor84, line 257: Instead of setting and then removing the label, just save in some bool variable if RoslynAnalyzer label should be added and then in the end just SetLabels once to proper value depending on that bool.

src/NuGetForUnity/Editor/NugetAssetPostprocessor.cs
JoC0de: Instead of `.Split(Path.DirectorySeparatorChar).First()` just use `Path.GetDirectoryName()` so platform specific stuff is handled correctly.

src/NuGetForUnity/Editor/NugetAssetPostprocessor.cs
```diff
@@ -180,7 +196,21 @@ private static string GetNuGetRepositoryPath()
...
+            {
+                return null;
+            }
+
+            return analyzerAssetPath.Substring(versionPrefixIndex + AnalyzersRoslynVersionSubFolderPrefix.Length)
+                .Split(Path.DirectorySeparatorChar)
+                .First();
+        }
+
+        private static void ModifyImportSettingsOfRoslynAnalyzer([NotNull] PluginImporter plugin, string[] relativeRepositoryPathComponents, bool reimport)
```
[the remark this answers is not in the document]
hadashiA: My mistake ><  Fixed.

src/NuGetForUnity/Editor/NugetAssetPostprocessor.cs
```diff
@@ -163,6 +173,12 @@ private static string[] GetPathComponents([NotNull] string path)
             return path.Split(Path.DirectorySeparatorChar);
         }
 
+        [NotNull]
+        private static string JoinPathComponents([NotNull] params string[] pathComponents)
```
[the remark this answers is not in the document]
hadashiA: I was mistaken. fixed it to use Path.Combine()

src/NuGetForUnity/Editor/NugetAssetPostprocessor.cs
```diff
@@ -180,7 +196,21 @@ private static string GetNuGetRepositoryPath()
...
+        {
+            var versionPrefixIndex = analyzerAssetPath.IndexOf(AnalyzersRoslynVersionSubFolderPrefix, StringComparison.Ordinal);
+            if (versionPrefixIndex < 0)
+            {
+                return null;
+            }
+
+            return analyzerAssetPath.Substring(versionPrefixIndex + AnalyzersRoslynVersionSubFolderPrefix.Length)
+                .Split(Path.DirectorySeparatorChar)
+                .First();
```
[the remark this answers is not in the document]
hadashiA: 👀 Based on your suggestion, I fixed it to a low-cost way.

src/NuGetForUnity/Editor/NugetAssetPostprocessor.cs
hadashiA: Ah this is a point of concern for me as well.

I installed the Nuget package on hand and tried it. (2021.3.16f, 2022.2.1f)

GetSubFolders seems to recognize folders before import is complete. 

Other methods I can think of are
- Directory.GetDirectories to query the file system directly.
- Rewrite the label in another Roslyn version.

However, I think GetSubFolders is simpler to implement. What do you think?

src/NuGetForUnity/Editor/NugetAssetPostprocessor.cs
hadashiA: `GetDirectoryName` behaves a little differently because it returns all path components before the file name.
(I want to get only the first path component.)

I have adopted the following comment suggestions. What do you think?
https://github.com/GlitchEnzo/NuGetForUnity/pull/616#discussion_r1457466067

src/NuGetForUnity/Editor/NugetAssetPostprocessor.cs
hadashiA: Ok, I fixed.

src/NuGetForUnity/Editor/NugetAssetPostprocessor.cs
igor84: You are still missing one thing to make this work. At the begging of the method add `var assetPath = Path.GetFullPath(plugin.assetPath)` and then user `assetPath` instead of `plugin.assetPath` bellow.

hadashiA: Oh, I understand. Fixed it.

src/NuGetForUnity/Editor/NugetAssetPostprocessor.cs
igor84: If you test that it works correctly in these cases that is good enough for me:
1. Right after installing a package from Unity
2. Turn off Unity, delete the installed package but leave it in packages.json and start Unity and confirm it is automatically installed correctly on startup.
3. Turn off Unity, delete the installed package but leave it in packages.json and also delete the Library folder and then start Unity and confirm it is automatically installed correctly on startup.
4. If there are multiple installed packages repeat the previous step but delete them all while leaving them in packages.json.

igor84 (approved):

src/NuGetForUnity/Editor/NugetAssetPostprocessor.cs
hadashiA: @igor84 I did the above test.

> Turn off Unity, delete the installed package but leave it in packages.json and start Unity

In this case, `AssetDatabase.GetSubFolders` does not seem to work, as you have pointed out.

So I rewrote the method to use `Directory.GetDirectories`. https://github.com/GlitchEnzo/NuGetForUnity/pull/616/commits/638c10a49e5bd8fc76849c1b272d25d0500caf98

With this commit, all steps 1-4 have passed.
Please check it out. Thanks.

popara96 (approved):

src/NuGetForUnity/Editor/NugetAssetPostprocessor.cs
JoC0de: Sorry for the delay.
The Path class also has function to get the last component of a path, the `GetFileName`. Sounds strange but it also works for directories. In your case you can combine GetDirectoryName and GetFileName. Both have versions that accept Span so you can avoid the allocation of the string.

hadashiA: @JoC0de I thought about it, but neither GetDirectoryName nor GetFileName seem to suit the purpose in the current implementation.
What I wanted to do is "get the first path component".


The current implementation is as follows:

1. The path of the Input is as:
    - `"/path/to/Assets/Packages/System.Text.Json.7.0.1/analyzers/dotnet/roslyn3.11/cs/System.Text.Json.SourceGeneration.dll"`
2. Check for the following patterns in the input:
    - `analyzers/dotnet/roslyn`
3. If a match is found, "3.11" is extracted. This is currently the following process.
    1. Get `analyzers/dotnet/roslyn` or later. 
        - -> `3.11/cs/System.Text.Json.SourceGeneration.dll"`
    2. Get the first component of the path separator.
        - -> `3.11`

src/NuGetForUnity/Editor/NugetAssetPostprocessor.cs
```diff
@@ -31,11 +31,32 @@ namespace NugetForUnity
     /// </summary>
     public class NugetAssetPostprocessor : AssetPostprocessor
     {
+        private const string MaxSupportedRoslynVersion =
+#if UNITY_2022_3_OR_NEWER && !UNITY_2022_3_1 && !UNITY_2022_3_2 && !UNITY_2022_3_3 && !UNITY_2022_3_3 && !UNITY_2022_3_4 && !UNITY_2022_3_5 && !UNITY_2022_3_6 && !UNITY_2022_3_7 && !UNITY_2022_3_8 && !UNITY_2022_3_9 && !UNITY_2022_3_10 && !UNITY_2022_3_11
```
JoC0de, line 35: We can't use `#if` toggles because the plugin is also released as a pre-build `.dll` so all compiler directives are already evaluated before imported into Unity.
Probably we should reuse the `UnityVersion` class from `src/NuGetForUnity/Editor/TargetFrameworkResolver.cs` witch would also make the statement a little bit easier.
Or can't we even detect the version using reflection with something like `typeof(Microsoft.CodeAnalysis.CSharp).Assembly.GetName().Version` (not testet) but as Unity has the "package" it should be possible to find the version of the assembly.

src/NuGetForUnity/Editor/NugetAssetPostprocessor.cs
```diff
@@ -189,7 +225,29 @@ private static void ModifyImportSettingsOfRoslynAnalyzer([NotNull] PluginImporte
...
+            // In that case, for the same package, the most recent version must be chosen out of those available for the current Unity version.
+            var assetPath = Path.GetFullPath(plugin.assetPath);
+            var assetRoslynVersion = GetRoslynVersionNumberFromAnalyzerPath(assetPath);
+            if (assetRoslynVersion != null)
+            {
+                var versionPrefixIndex = assetPath.IndexOf(AnalyzersRoslynVersionsFolderName, StringComparison.Ordinal);
+                var analyzersVersionsRoot = Path.Combine(assetPath.Substring(0, versionPrefixIndex), AnalyzersRoslynVersionsFolderName);
+                var analyzersFolders = Directory.GetDirectories(analyzersVersionsRoot);
+                var enabledRoslynVersions = analyzersFolders.Select(GetRoslynVersionNumberFromAnalyzerPath)
+                    .Where(version => version != null && string.CompareOrdinal(version, MaxSupportedRoslynVersion) <= 0)
```
JoC0de, line 240: Comparing version number strings using 'string comparison' doesn't work in all cases, e.g. `1.10.0` will be lower than `1.9.0`. You can use the `NugetPackageVersion` class it supports comparison of version numbers including pre-release versions etc.

JoC0de (changes requested):
Thanks for the contribution.
I still don't understand why Unity doesn't handle this by itself but at least we can fix it 😃

src/NuGetForUnity/Editor/NugetAssetPostprocessor.cs
hadashiA: Ok, I fixed it to use NugetPackageVersion.

src/NuGetForUnity/Editor/NugetAssetPostprocessor.cs
hadashiA: @JoC0de I fixed `UnityVersion` to be internal instead of private and use it.

> We can't use #if toggles because the plugin is also released as a pre-build .dll so all compiler directives are already evaluated before imported into Unity.
> Probably we should reuse the UnityVersion class from src/NuGetForUnity/Editor/TargetFrameworkResolver.cs witch would also make the statement a little bit easier.

As far as I know, NugetForUnity is provided as a Unity Package with the source code, so #if seems to work too.
But as you say, it would make sense to share the UnityVersion logic. I did so.

> Or can't we even detect the version using reflection with something like typeof(Microsoft.CodeAnalysis.CSharp).Assembly.GetName().Version (not testet) but as Unity has the "package" it should be possible to find the version of the assembly.

`Microsoft.CodeAnalysis.CSharp` is not linked to the Editor or Runtime context of Unity. And it  is not present in the packages our Unity application depends on.
It is used by the Unity system to compile the application layer.
I don't know how to access this at runtime from the Editor context. Any idea?

JoC0de (approved):

PR #617, opened 2024-01-21
Improve GUI of NuGet for Unity window

AnnulusGames:
NuGet for Unity is a great tool, but its GUI looks a little outdated and feels clunky. This pull request will refresh the look of the editor window and bring it in line with the modern Unity editor.

### Online

img1-before
img1-after

### Installed

img2-before
img2-after

### Updates

img3-before
img3-after


JoC0de (approved):
@AnnulusGames Thank you for the pull request! Your contribution is greatly appreciated and enhances the project. Nice work!
I run autoformatter on the changes and added the missing documentation, so from my side we can merge it. Is it still work in progress or can we merge it?

AnnulusGames:
I've updated the images in the documentation. The work has already been completed, so there is no problem in merging it.

src/NuGetForUnity/Editor/Ui/NugetWindow.cs
```diff
@@ -711,88 +594,88 @@ private void DrawPackagesSplittedByManuallyInstalled(List<INugetPackage> package
...
             {
-                var arrow = !showPackagesToInstall ? ArrowTipUp : ArrowTipDown;
-                if (GUILayout.Button($" {arrow} Online packages", headerStyle, GUILayout.Height(25f)))
-                {
-                    showPackagesToInstall = !showPackagesToInstall;
-                }
+                DrawPackages(availablePackages.Where(p => !selectedPackageInstalls.Contains(p)), true);
             }
 
+            EditorGUILayout.BeginVertical();
```
igor84, line 631: Why do you need this BeginVertical? Aren't we already in vertical scope here?

src/NuGetForUnity/Editor/Ui/NugetWindow.cs
```diff
@@ -806,31 +689,18 @@ private void DrawSelectedForInstallationHeader(GUIStyle headerStyle)
...
         {
-            var backgroundStyle = GetBackgroundStyle();
-            var contrastStyle = GetContrastStyle();
+            var backgroundStyle = Styles.BackgroundStyle;
 
             foreach (var package in packages)
             {
                 using (new EditorGUILayout.VerticalScope(backgroundStyle))
                 {
-                    DrawPackage(package, backgroundStyle, contrastStyle, canBeSelected);
```
igor84, line 829: I actually liked the stripped look but I guess it looks ok like this too :D.

src/NuGetForUnity/Editor/Ui/NugetWindow.cs
```diff
@@ -839,59 +709,62 @@ private void DrawPackages(IEnumerable<INugetPackage> packages, bool canBeSelecte
...
                     {
                         // the search button emulates the Enter key
                         enterPressed = true;
                     }
 
-                    GUI.skin.textField.fontSize = oldFontSize;
+                    // search only if the enter key is pressed
+                    if (enterPressed)
+                    {
+                        GUI.FocusControl(string.Empty);
```
igor84, line 751: Why do you deselect the search box? It seems it would be useful to modify the search term if you search didn't provide satisfactory results so you can try again.

src/NuGetForUnity/Editor/Ui/NugetWindow.cs
```diff
@@ -839,59 +709,62 @@ private void DrawPackages(IEnumerable<INugetPackage> packages, bool canBeSelecte
...
                     DrawSelectFromClipboardButton();
                     DrawMandatoryButtons();
                 }
 
                 EditorGUILayout.EndHorizontal();
 
-                var enterPressed = Event.current.Equals(Event.KeyboardEvent("return"));
-
-                EditorGUILayout.BeginHorizontal();
+                var style = new GUIStyle(EditorStyles.toolbar) { fixedHeight = 25f };
```
igor84, line 734: This style is used in multiple places. Why not cache it in a field and reuse it instead of creating it all the time?

src/NuGetForUnity/Editor/Ui/NugetWindow.cs
```diff
@@ -1146,30 +1024,11 @@ private void DrawPackage(INugetPackage package, GUIStyle packageStyle, GUIStyle
                     // text is allowed to get the half of the available space rest is for buttons and version label
                     rect.width = (position.width - rect.x) / 2;
 
-                    EditorStyles.label.fontStyle = FontStyle.Bold;
-                    EditorStyles.label.fontSize = 16;
+                    var labelStyle = new GUIStyle(EditorStyles.label) { fontStyle = FontStyle.Bold, fontSize = 15 };
```
igor84, line 1027: Since this method is called in a loop for each package you should probably cache this style in a field.

igor84 (approved):

src/NuGetForUnity/Editor/Ui/NugetWindow.cs
AnnulusGames: This was unnecessary, so I removed it.

src/NuGetForUnity/Editor/Ui/NugetWindow.cs
AnnulusGames: This is a piece of code I wrote when I tried to remove the Search button, but it ended up being unnecessary as I kept the button. Therefore, I removed it.

src/NuGetForUnity/Editor/Ui/NugetWindow.cs
AnnulusGames: I cached it inside the Styles class.

src/NuGetForUnity/Editor/Ui/NugetWindow.cs
```diff
@@ -987,34 +857,39 @@ private void DrawInstalledHeader()
         /// </summary>
         private void DrawUpdatesHeader()
         {
-            var headerStyle = GetHeaderStyle();
+            var headerStyle = Styles.HeaderStyle;
 
             EditorGUILayout.BeginVertical(headerStyle);
```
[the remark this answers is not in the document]
AnnulusGames: I replaced `BeginHorizontal/Vertical()` with `using(...)` for the most part. Some parts are left as they are because I thought that `BeginHorizontal/Vertical()` would be better in that part.

src/NuGetForUnity/Editor/Ui/NugetWindow.cs
AnnulusGames: Like other styles, I cached it in the Styles class.

src/NuGetForUnity/Editor/Ui/NugetWindow.cs
```diff
@@ -1290,14 +1147,42 @@ private void DrawPackage(INugetPackage package, GUIStyle packageStyle, GUIStyle
...
+            GUILayout.Space(7f);
+
             using (new EditorGUILayout.HorizontalScope())
             {
                 using (new EditorGUILayout.VerticalScope())
                 {
                     // Show the package details
-                    EditorStyles.label.wordWrap = true;
-                    EditorStyles.label.fontStyle = FontStyle.Normal;
+                    var labelStyle = new GUIStyle(EditorStyles.label)
```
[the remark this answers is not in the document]
AnnulusGames: I cached this as well.

AnnulusGames:
I think the best way is to replace the drawing process from IMGUI to UIToolkit, but it would be difficult to migrate this huge amount of code to it, so I improved the appearance by rewriting the existing code. I think this is enough for now.

PR #618, opened 2024-01-21
Option to install in /Packages/NuGetAssets

Akeit0:
Add an option to install not to  Assets Folder but, Packages/NuGetAssets Folder 
スクリーンショット 2024-01-21 175748

### DIFF
--- a/src/NuGetForUnity/Editor/Configuration/NugetConfigFile.cs
+++ b/src/NuGetForUnity/Editor/Configuration/NugetConfigFile.cs
@@ -87,25 +87,24 @@ namespace NugetForUnity.Configuration
         ///     The underlying absolute path where packages are to be installed.
         /// </summary>
         [NotNull]
-        private string underlyingRepositoryPath= Path.GetFullPath(Path.Combine(Application.dataPath, "Packages"));
+        private string underlyingRepositoryPath { get; set; } = Path.GetFullPath(Path.Combine(Application.dataPath, "Packages"));
         
         /// <summary>
         ///     Gets the absolute path where packages are to be installed.
         /// </summary>
         [NotNull]
-        public string RepositoryPath { get=>underlyingRepositoryPath;
-             set => underlyingRepositoryPath = savedRepositoryPath = value;
+        public string RepositoryPath 
+        { 
+            get=>underlyingRepositoryPath;
+            set => underlyingRepositoryPath = savedRepositoryPath = value;
         }
 
         /// <summary>
         ///     Gets the relative path to directory where packages are to be installed. The path is relative to the folder containing the 'NuGet.config' file.
         /// </summary>
         [NotNull]
-        public string RelativeRepositoryPath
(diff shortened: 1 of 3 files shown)


igor84:
Thank you for your effort but I don't think we can accept this pull request. Disregarding all the formatting errors and a bit strange implementation of property in UnityPathHelper I think this would clash with a functionality we already have. You see if the package that is being installed contains a `tools` folder the contents of just that folder will be moved to Packages folder while the rest of the files, if any, will be installed somewhere under Assets directory.  With your change installation of such packages would fail with strange errors.

Also there is not much sense in managed packages that are not used within Unity with NugetForUnity. Are you sure you can't just use the existing `tools` functionality for your use case?

Akeit0:
スクリーンショット 2024-01-23 113030
I changed implementation to one that allows config.RepositoryPath to be changed from Preferences.
Also, folders are now moved automatically.
(Editor needs to be restarted when moving dll's, but this is probably unavoidable)

I think this won't clash with a current functionality, because these are only what was originally possible with a text editor.

> You see if the package that is being installed contains a tools folder the contents of just that folder will be moved to Packages folder while the rest of the files, if any, will be installed somewhere under Assets directory. With your change installation of such packages would fail with strange errors.

Is this due to my change? I tried and no error occurs and it seems to be normal behavior for the tools folder to be located in the Packages folder.

Akeit0:
Simply because my acquaintance and I are disturbed that Packages is in Assets folder. 
I think it is a waste not to use the RepositoryPath setting that is already there.

igor84:
But I don't understand what is the point of installing packages for Unity in a folder where Unity will not see them and will not be able to use them? I tried just moving my existing packages to that folder and I just got a lot of compile errors. As far as I know Unity can only see packages that are specified in its Packages/manifest.json file from that folder.

Akeit0:
You can see https://github.com/GlitchEnzo/NuGetForUnity/pull/618#issue-2092436116 , how it actually works.

igor84:
Ah, ok. I investigated it a bit more and I now understand how is this supposed to work. There are still a few more cases that needs to be handled, like a use case where users want to gitignore the folder where the packages are installed but still have everything resolve automatically after Unity start and I think some path handling will not be correct as it is. I'll try to find some time to go through this, maybe create a separate pull requests where I try to cover all the cases...

igor84:
I opened a pull request #619 where I tried to cover this. Can you please review if that is it and tell me if you are ok that we go with that pull request instead of this one.

Akeit0:
> I opened a pull request #619 where I tried to cover this. Can you please review if that is it and tell me if you are ok that we go with that pull request instead of this one.

Thank you! You did a good job!
But isn't it somewhat unnatural that selecting the /Package folder generates a folder under it?
That's the only thing that bothered me.

Akeit0:
It would be better to be more explicit or display a note in Preferences.

Issue #495, opened 2023-02-06
Doesn't work with a proxy


wightwhale:
I can't get this plugin to work behind a proxy. I've tried changing the config files to use the proxy but it doesn't seem to make a difference. Visual Studio has a popup that lets you enter username and password to bypass the proxy.

wightwhale:
I receive the following error when trying to use the plugin through the proxy here at work.
Curl error 56: Received HTTP code 407 from proxy after CONNECT

I have the http_proxy and https_proxy environment variables set and they work with the unity package manager. The proxy here requires a user and pw to bypass it. ex. http://username:password@proxy.companywebsite.com:8080

PR #620, opened 2024-01-28
Add basic proxy support for NuGet API V3

k6a:
* addresses proxy connection issue #495

The default proxy setup was missing in HttpClient/HttpClientHander initializaion.
By adding WebRequest.DefaultWebProxy client will look for HTTP_PROXY, NO_PROXY environment variables if it exists.

Tested on Unity 2018.3.0, 2022.3.18, and Squid 5.7.

### DIFF
--- a/src/NuGetForUnity/Editor/PackageSource/NugetApiClientV3.cs
+++ b/src/NuGetForUnity/Editor/PackageSource/NugetApiClientV3.cs
@@ -44,14 +44,7 @@ namespace NugetForUnity.PackageSource
 
         [NonSerialized]
         [NotNull]
-        private readonly HttpClient httpClient = new HttpClient(
-            new HttpClientHandler
-            {
-                AutomaticDecompression = DecompressionMethods.GZip | DecompressionMethods.Deflate,
-#if UNITY_EDITOR_WIN
-                Proxy = WebRequest.GetSystemWebProxy(),
-#endif
-            });
+        private readonly HttpClient httpClient;
 
         [NonSerialized]
         [CanBeNull]
@@ -90,6 +83,23 @@ namespace NugetForUnity.PackageSource
             apiIndexJsonUrl = new Uri(url);
 
             InitializeFromSessionState();
+
+            // On Windows, Mono HttpClient does not automatically pick up proxy settings.
+            if (Application.platform == RuntimePlatform.WindowsEditor)
+            {
+                httpClient = new HttpClient(new HttpClientHandler
+                {
+                    AutomaticDecompression = DecompressionMethods.GZip | DecompressionMethods.Deflate,
+                    Proxy = WebRequest.GetSystemWebProxy()
+                });
+            }
+            else
+            {
+                httpClient = new HttpClient(new HttpClientHandler
+                {
+                    AutomaticDecompression = DecompressionMethods.GZip | DecompressionMethods.Deflate
+                });
+            }
         }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


JoC0de:
Are you sure it didn't work as it was before? As far as I understand the documentation https://learn.microsoft.com/en-us/dotnet/api/system.net.http.httpclienthandler.proxy?view=net-8.0 it is the default behavior.

k6a:
@JoC0de Thanks for your response!

I just got started using this great plugin in about a week ago. Still figuring out how to use it in corporate proxy (harsh) environment.

Further tested and found that the symptom only occurs on Windows + (corporate) proxy enabled environment.

On MacOS, it works just fine (without the fix).

The documentation you mentioned explains the behaviour of HttpClientHandler on **.NET runtime**.

Since the Unity Editor runs on **Mono runtime**, internal structures and behaviours are slightly different than that of .NET runtime.

Reference source for both runtimes:

.NET: 

* https://github.com/dotnet/runtime/blob/main/src/libraries/System.Net.Http/src/System/Net/Http/HttpClient.cs
* https://github.com/dotnet/runtime/blob/main/src/libraries/System.Net.Http/src/System/Net/Http/HttpClientHandler.cs

Mono:

* https://github.com/mono/mono/blob/main/mcs/class/System.Net.Http/System.Net.Http/HttpClient.cs
* https://github.com/mono/mono/blob/main/mcs/class/System.Net.Http/HttpClientHandler.cs
 
And yes, .NET does setup system proxy on initialization whereas Mono doesn't.

src/NuGetForUnity/Editor/PackageSource/NugetApiClientV3.cs
```diff
@@ -45,7 +45,13 @@ internal sealed class NugetApiClientV3 : IDisposable
         [NonSerialized]
         [NotNull]
         private readonly HttpClient httpClient = new HttpClient(
-            new HttpClientHandler { AutomaticDecompression = DecompressionMethods.GZip | DecompressionMethods.Deflate });
+            new HttpClientHandler
+            {
+                AutomaticDecompression = DecompressionMethods.GZip | DecompressionMethods.Deflate,
+#if UNITY_EDITOR_WIN
```
JoC0de, line 51: Can we enable it on all platforms?
We can't use `#if` toggles because the plugin is also released as a pre-build `.dll` so all compiler directives are already evaluated. If we need this to be platform specific we need to use e.g. `if (Application.platform == RuntimePlatform.WindowsEditor)`

JoC0de (changes requested):

src/NuGetForUnity/Editor/PackageSource/NugetApiClientV3.cs
k6a: @JoC0de 

Thanks for pointing out about the pre-build .dll behaviour. This really helps me.
Upon regression testing, I found that the fix introduces negative effects on MacOS platform. 
So the fix cannot be enabled on all platforms. This is Windows only fix/workaround.

I've updated and moved the HttpClient initialization code to the constructor.
Modified to use Applicaton.platform conditional check as you advised.

Test method:

* Open NuGetForUnity window
* Search NuGet package with some dependencies
* Install NuGet package
* Unintall NuGet package

Test results:

1. DLL version

| Network Environment | Windows | MacOS |
| - | - | - |
| Non Proxy | 20240210_OK | 20240210_OK |
| Proxy | 20240210_OK | 20240210_OK |

2. Script version

| Network Environment | Windows | MacOS |
| - | - | - |
| Non Proxy | 20240210_OK | 20240210_OK |
| Proxy | 20240210_OK | 20240210_OK |

JoC0de (approved):

src/NuGetForUnity/Editor/PackageSource/NugetApiClientV3.cs
JoC0de: Ok thanks for the effort of testing it on the devices.

PR #631, opened 2024-03-11
Support for native dependencies

JoC0de:
Rebased version of #421
with some additional changes
image

I still work on it / it still needs some small changes.


JoC0de:
@igor84 @popara96 I think it is now mergeable. Probably the warning message should recommend testing on device or somehow note that there can be problems. Currently I have no good formulation especially as Unity clamps the message on newer versions, so it can't be much text.

igor84 (approved):
Looks good. Btw. why can't we also support arm64 for Android and iOS?

Issue #621, opened 2024-02-02
.NET Framework is prioritized over .NET Standard 2.1 as a target framework.


mayuki:
## Description

Starting with Unity 2021.2, .NET Standard 2.1 is supported. In those versions, if you set the `API Compatibility Level` to `.NET Framework`, the target framework selected in the NuGet package will prioritize .NET Framework over .NET Standard 2.1.

For example, if the package includes `net462` and `netstandard2.1`, and you set the `API Compatibility Level` to `.NET Framework` in Unity 2021.3, `net462` will be chosen.

In the original .NET, this selection would be correct. However, the situation differs in Unity because Unity's .NET Framework includes .NET Standard 2.1. Unity's `.NET Framework` represents an API set: `.NET Framework 4.x + .NET Standard 2.1`.

https://docs.unity3d.com/2021.3/Documentation/Manual/dotnetProfileSupport.html

In cases where a package includes both `net462` and `netstandard2.1`, it is usually assumed that the implementation of `.NET Standard 2.1` is more modern. When the `API Compatibility Level` is set to `.NET Standard`, NuGetForUnity selects `netstandard2.1`, but when changed to `.NET Framework`, the more legacy `net462` is selected, leading to potential problems.

This is not just an issue of legacy. While .NET Standard is platform-neutral, .NET Framework targets the Windows platform, which could lead to non-functionality or unnecessary dependencies being installed. Grpc.Net.Client works with `netstandard2.1`, but it does not function properly when `net462` is selected.

-   **NuGet Package:** Grpc.Net.Client
-   **NuGetForUnity Version:** 4.0.2
-   **Unity Version:** Unity 2022.3
-   **Operating System:** Windows

Reag:
Id like to +1 this issue. As it stands right now, my automated pipeline cannot function without adding all the NuGet Dlls to my repo manually. This goes directly against the main purpose of NuGet, not having to store all these dependencies in your own source control.

Put simply, this issue is a showstopper for using this tool. As long as its getting net461 versions of libraries that will not compile into unity, I cannot use this for its intended purpose.

PR #641, opened 2024-04-26
Add option to prefer .NET Standard over .NET Framework as TargetFramework

mayuki:
This PR adds the option to prioritize .NET Standard over .NET Framework as TargetFramework.

image

## Background
As written in Issue https://github.com/GlitchEnzo/NuGetForUnity/issues/621, the currently supported Unity differs from the original .NET Framework, possessing an API surface of .NET Standard 2.1, even when the API Compatibility Level is set to `.NET Framework`.

Due to [Unity's .NET Framework a strange and unique combination of .NET Framework 4.8 and .NET Standard 2.1](https://docs.unity3d.com/2022.3/Documentation/Manual/dotnetProfileSupport.html), various issues arise when an assembly targeted at `net46x` is installed.

For example, `Microsoft.Bcl.AsyncInterfaces` for `net462` (.NET Framework) includes `System.IAsyncDisposable`, which is provided as an in-built API in .NET Standard 2.1. As a result, installing it in a .NET Framework Unity project causes the following compile error:

```
error CS0433: The type 'IAsyncDisposable' exists in both 'Microsoft.Bcl.AsyncInterfaces, Version=8.0.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51' and 'mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089'
```

Additionally, `Grpc.Net.Client` has libraries for `net461` that are not compatible with .NET Standard 2.x and do not work, leading to unnecessary dependencies being installed.

To address this issue, https://github.com/GlitchEnzo/NuGetForUnity/pull/630 made it possible to specify `targetFramework` in the packages.config, but this does not affect transitive dependencies. In some cases, you need to specify this for each implicitly installed package. (Example: [Our project's installation of Grpc.Net.Client in package.json](https://github.com/Cysharp/MagicOnion/blob/da3b2db502a544e9afbc728c5fc47b9f9ef30b53/samples/ChatApp/ChatApp.Unity/Assets/packages.config))

Hence, I added an option to always prioritize .NET Standard 2.x over .NET Framework as the target framework choice in the project.

## Remarks
<del>The option added by this PR is **turned off by default**, but in my opinion, making this behavior of prioritizing .NET Standard the default could be appropriate.</del>

The option added by this PR is **turned off by default**. This means that the behaviour does not change in existing projects. But it **turned on by default** when NuGet.config is newly created.

I believe that libraries built for .NET Standard 2.x, especially 2.1, are more modern than those built specifically for the .NET Framework. In Unity, if `API Compatibility Level` changes from `.NET Standard` to `.NET Framework`, the legacy API set could be used instead of the libraries that could be used in .NET Standard 2.1, but I don't think any developer would like that.

### DIFF
--- a/src/NuGetForUnity.Tests/Assets/Tests/Editor/NuGetTests.cs
+++ b/src/NuGetForUnity.Tests/Assets/Tests/Editor/NuGetTests.cs
@@ -816,6 +816,66 @@ public class NuGetTests
     }
 
     [Test]
+    [TestCase(
+        "2021.3.16f1",
+        true,
+        new[] { "net4.7.2", "netstandard2.0", "unity", "netstandard2.1" },
+        new[] { "unity", "netstandard2.1", "netstandard2.0", "net4.7.2" })]
+    [TestCase(
+        "2021.3.16f1",
+        false,
+        new[] { "net4.7.2", "netstandard2.0", "unity", "netstandard2.1" },
+        new[] { "unity", "net4.7.2", "netstandard2.1", "netstandard2.0" })]
+    public void TryGetBestTargetFrameworkPreferNetStandardTest(
+        string unityVersion,
+        bool preferNetStandard,
+        string[] availableFrameworks,
+        string[] expectedBestMatch)
+    {
+        var unityVersionType = typeof(UnityVersion);
+        var currentUnityVersionProperty = unityVersionType.GetProperty(nameof(UnityVersion.Current), BindingFlags.Public | BindingFlags.Static);
+        Assume.That(currentUnityVersionProperty, Is.Not.Null);
+        Assume.That(currentUnityVersionProperty.CanRead, Is.True);
+        Assume.That(currentUnityVersionProperty.CanWrite, Is.True);
+
+        var currentBuildTargetApiCompatibilityLevelProperty = typeof(TargetFrameworkResolver).GetProperty(
+            nameof(TargetFrameworkResolver.CurrentBuildTargetApiCompatibilityLevel),
+            BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Static);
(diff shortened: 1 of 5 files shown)


igor84 (approved):
Nice, clean work with great explanation :)

mayuki:
Thanks for the quick review!

By the way, what do you think about enabling it by default?

I believe that having a more modern API set (`.NET Standard 2.1`) as the primary choice, even if `.NET Framework` is selected as `API Compatibility Level`, is not a bad idea for many developers, but it will definitely introduce breaking changes, so I would like to hear your opinions.

igor84:
I never used net standard packages in net framework Unity so I can't speak from experience. We actually switched all our Unity projects to net standard before we started using nuget and never had a need to switch back.

Anyway we usually handle things like this by leaving the new option as default false in code but we change the template for new nuget.config to set it to true. That way the option becomes turned on by default only for new projects while not breaking the existing ones.

mayuki:
I initially thought that .NET Standard would be sufficient in the modern scenario, but it seems that more developers than expected are still using .NET Framework and I have received feedback about various problems.

This choice is often forced by third-party SDKs and middleware that require .NET framework, irrespective of the developer's preferences.

Now I will change it so that this option is **enabled by default in new `nuget.config` template** and **disabled by default in the code**.

JoC0de:
@mayuki @igor84 Sorry for the delay. I simplified the creation of the `PrioritizedTargetFrameworksPreferNetStandard20Or21`, added a unit test and changed the UI a little bit. 
Would be nice if you can have a quick look at the changes and 'approve' them, than we can merge the PR.

igor84 (approved):

PR #658, opened 2024-06-21
Fix duplicated files in folder content and contentFiles

cyanial:
https://stackoverflow.com/questions/47468941/prevent-duplicating-files-in-nuget-content-and-contentfiles-folders

Test NuGetPackage: `MongoDB.Libmongocrypt`.

as far as i know, that a nuget package may have both `content` and `contentFiles` folder and the contain the same files.

from now on, i change the code to omit `content` folder when it has `contentFiles` and it works for my package.

### DIFF
--- a/src/NuGetForUnity/Editor/NugetPackageInstaller.cs
+++ b/src/NuGetForUnity/Editor/NugetPackageInstaller.cs
@@ -171,7 +171,7 @@ namespace NugetForUnity
                         var hasContentFilesFolder = false;
                         foreach (var entry in zip.Entries)
                         {
-                            if (entry.FullName.EndsWith("/") && entry.FullName.Equals(contentFilesDirectoryName, StringComparison.OrdinalIgnoreCase))
+                            if (entry.FullName.StartsWith(contentFilesDirectoryName, StringComparison.Ordinal))
                             {
                                 hasContentFilesFolder = true;
                                 break;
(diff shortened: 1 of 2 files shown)


src/NuGetForUnity/Editor/NugetPackageInstaller.cs
```diff
@@ -166,6 +166,18 @@ private static bool Install([NotNull] INugetPackage package, bool refreshAssets,
...
+                        const string contentFilesDirectoryName = "contentFiles/";
+                        var hasContentFilesFolder = false;
+                        foreach (var entry in zip.Entries)
+                        {
+                            if (entry.FullName.StartsWith(contentFilesDirectoryName, StringComparison.Ordinal))
+                            {
+                                hasContentFilesFolder = true;
+                                break;
+                            }
+                        }
```
JoC0de, line 179:
```suggestion
                        var hasContentFilesFolder = zip.Entries.Any(entry => entry.FullName.StartsWith(contentFilesDirectoryName, StringComparison.Ordinal));
```

JoC0de (changes requested):

JoC0de:
Sorry for the delay.
Thanks for the workaround.
Would be nice if we add a UnitTest, I could implement it but not sure when I have time.

cyanial:
> Sorry for the delay. Thanks for the workaround. Would be nice if we add a UnitTest, I could implement it but not sure when I have time.

never mind, i'm now using my own version, it works. 

And i'm studying this project as well, i'm also trying to add some unit test. but i'm a new starter in C#, so forgive for my poor c# skills.

cyanial (commented):
👍

PR #696, opened 2025-01-30
Made RemoveUnnecessaryPackages consistent

igor84:
In the old code it would look for nuspec files on the file system and then remove folders that are not in packages.config. But installedPackages already contain all found nuspec files. The issue with old code is that InstalledPackages contained data about packages that were also potentially modified by the plugins while this method didn't take that into account leading to inconsistencies.

### DIFF
--- a/src/NuGetForUnity/Editor/InstalledPackagesManager.cs
+++ b/src/NuGetForUnity/Editor/InstalledPackagesManager.cs
@@ -256,38 +256,17 @@ namespace NugetForUnity
                 return false;
             }
 
-            var directories = Directory.GetDirectories(ConfigurationManager.NugetConfigFile.RepositoryPath, "*", SearchOption.TopDirectoryOnly);
             var somethingDeleted = false;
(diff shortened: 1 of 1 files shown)


src/NuGetForUnity/Editor/InstalledPackagesManager.cs
```diff
@@ -256,38 +256,17 @@ internal static bool RemoveUnnecessaryPackages()
                 return false;
             }
 
-            var directories = Directory.GetDirectories(ConfigurationManager.NugetConfigFile.RepositoryPath, "*", SearchOption.TopDirectoryOnly);
```
JoC0de, line 259: pleas adjust the documentation of the `RemoveUnnecessaryPackages` method as it now doesn't check for a directory.

JoC0de (changes requested):

popara96 (approved):

JoC0de (approved):

Issue #697, opened 2025-02-12
Add capability to expend environment variables in username - for packageSourceCredentials


QuentinBrun:
Hi,

As it is done for password, that would be great to be able to expend variables in the username field. Use case example include user name change between developper, but also for CI/CD.

Also variable expension in the username is handled in native nuget.

If needed, I may be able to propose a quick implementation.

If there is another way to configure user name, please make me aware.

Thanks

JoC0de:
As a alternative you can use System wide `NuGet.config` file introduced in: https://github.com/GlitchEnzo/NuGetForUnity/pull/678

QuentinBrun:
Thank you for doing it so quickly I appreciate.

PR #698, opened 2025-02-12
#697 add capability to expend env variable in username

QuentinBrun:
I hope this help.


JoC0de:
I am not sure if we can implement it in this easy way. The `UserName` property should be renamed to `SavedUserName` to be consistent with the `SavedPassword` property. But it affects the public interface `INugetPackageSource` witch impacts our plugin system.
@igor84 what is your opinion on this?

igor84:
Our plugin system is still very young so I would be surprised if anyone beside me is using it :) and since I don't use username and in my plugin and configuration it wouldn't really break anything... So I would be ok with changing it.

JoC0de (approved):

QuentinBrun:
Hi, thank you!

PR #699, opened 2025-02-15
Fixed Update All so it correctly updates implicit dependencies

igor84:
NugetWindow.updatePackages actually contains all installed packages with all the versions since we use it to quickly filter between Show Downgrades/Show Updates option. On the other hand Update All button is only visible when Show Updates option is active and so we need to pass the FilteredUpdatePackages to UpdateAll method since they will actually contain only  the packages with updates.


JoC0de:
@igor84 I adjusted it so it doesn't bother the current search term -> so it updates all packages that have a update but don't need to match the entered search text

JoC0de (approved):

Issue #709, opened 2025-03-31
When using GitHub Nuget feed, "restore" and "install" updates to latest version rather than the (older) version in packages.config


AlessandroCorrad:
Hi,
so the issue is the following:
when I try to "install" or "restore" a NuGet package that is hosted on GitHub, it always installs the latest version of the package rather than the (older) version that I specified via packages.config or via the Editor Window.
The package is hosted on a company private GitHub repository and it requires authentication via GitHub token. The issue is not present if I execute the same commands via the official dotnet CLI in a .Net empty project so I'm pretty sure the package is available in the registry.

Operating system: Windows 10
NuGetForUnity: 4.3.0

bigswede74:
I have the same issue see the below logs where the latest version is `0.1.143-prerelease` but the version I am looking for is `0.1.47-prerelease`.

```
Restoring NuGet Packages ...
Failed to get 'my.package' from the package source 'https://api.nuget.org/v3/index.json'. Error: System.Net.Http.HttpRequestException: The request to 'https://api.nuget.org/v3/registration5-gz-semver2/my.package/index.json' failed with status code 'NotFound' and message: <?xml version="1.0" encoding="utf-8"?><Error><Code>BlobNotFound</Code><Message>The specified blob does not exist.
RequestId:XXXXXXXXXXX
Time:2025-04-03T23:35:32.4027760Z</Message></Error>
   at NugetForUnity.PackageSource.NugetApiClientV3.EnsureResponseIsSuccessAsync(HttpResponseMessage response) in /_/src/NuGetForUnity/Editor/PackageSource/NugetApiClientV3.cs:line 458
   at NugetForUnity.PackageSource.NugetApiClientV3.GetStringFromServerAsync(NugetPackageSourceV3 packageSource, String url, CancellationToken cancellationToken) in /_/src/NuGetForUnity/Editor/PackageSource/NugetApiClientV3.cs:line 763
   at NugetForUnity.PackageSource.NugetApiClientV3.GetRegistrationPageItemsAsync(NugetPackageSourceV3 packageSource, INugetPackageIdentifier package, CancellationToken cancellationToken) in /_/src/NuGetForUnity/Editor/PackageSource/NugetApiClientV3.cs:line 545
Installing my.package 0.1.143-prerelease ...
my.package 0.1.47-prerelease is already listed in the packages.config file.  Updating to 0.1.143-prerelease
```

pocketninja25:
[...]

PR #711, opened 2025-04-11
Sort pageitems before searching for correct package version (#709)

pocketninja25:
Attempt to fix issue where some nuget packages would forcibly update themselves to a higher version when performing a restore on the command line. #709


bigswede74:
@JoC0de @GlitchEnzo , Small PR here, shouldn’t take more than a few minutes to review.

JoC0de (approved):

JoC0de:
@bigswede74 sorry for the delay. I merged it. Would be nice if you test if it fixes the issue. The GitHub action outputs the newest release as a `.unitypackage` or you can install it using the "Git" install method. Would be nice if I know if the fix works before releasing a new version.

pocketninja25:
This version does appear to fix the issue we were having. Succesfully tried it on a couple of projects

Issue #714, opened 2025-05-02
Updated Packages don't appear in the Updates Tab


TechGoon:
I'm using a custom Nuget server hosted through Github Packages to supply libraries to my unity project.  When I update a custom package, the option to update that package within unity does not appear, I must instead uninstall and reinstall the package.

Context:

NugetForUnity version: 4.3.0
Unity version: 2022.3.45f1
Configuration:
![Image](https://github.com/user-attachments/assets/285426da-f25f-4368-ad1e-81a6974c5cc7)

STR: 

With NugetForUnity configured to source packages from Nuget as well as a Custom Package Source:
1. Add an existing custom package to the project
2. Publish a new version of that package with an incremented version string.
4. In Unity, go to the Nuget Menu and choose "Manage Nuget Packages".
5. Note that a warning and an error appear in the console. (see below)
6. Go to the "Updates" tab

Expected Result: 
The new version of that package is listed on the Updates tab, and I can click a button to update that package

Actual Result: 
No packages appear on the Updates tab.

Warnings:
```
Failed to get 'Xxxx.Xxxxxxxx' from the package source 'https://api.nuget.org/v3/index.json'. Error: System.Net.Http.HttpRequestException: The request to 'https://api.nuget.org/v3/registration5-gz-semver2/Xxxx.Xxxxxxxx/index.json' failed with status code 'NotFound' and message: <?xml version="1.0" encoding="utf-8"?><Error><Code>BlobNotFound</Code><Message>The specified blob does not exist.
RequestId:9a19571f-801e-0041-2570-bbe513000000
Time:2025-05-02T14:45:20.5346542Z</Message></Error>
  at NugetForUnity.PackageSource.NugetApiClientV3.EnsureResponseIsSuccessAsync (System.Net.Http.HttpResponseMessage response) [0x000b0] in .\Library\PackageCache\com.github-glitchenzo.nugetforunity@fe31ad0904\Editor\PackageSource\NugetApiClientV3.cs:458 
  at NugetForUnity.PackageSource.NugetApiClientV3.GetStringFromServerAsync (NugetForUnity.PackageSource.NugetPackageSourceV3 packageSource, System.String url, System.Threading.CancellationToken cancellationToken) [0x0011e] in .\Library\PackageCache\com.github-glitchenzo.nugetforunity@fe31ad0904\Editor\PackageSource\NugetApiClientV3.cs:763 
[...]

PR #715, opened 2025-05-04
Fix update for combined V3 package source

JoC0de:
fix update when using multiple V3 sources (there could be a 'null' package added to the list if the package was not available on the package source)

fixes: #714


igor84:
Hi @JoC0de. We just hit this bug in our attempt to migrate from Artifactory to Gitea. Do you think you could release a new version soon with this fix?

PR #718, opened 2025-05-14
Sped up the refresh of package updates

igor84:
We added additional Artifactory source that has a lot of packages with a lot of versions and opening Nuget window started taking 20s. I moved fetching updates when window is opened to a background thread so that window is opened much faster. I also sped up the fetching of package updates by using Task.WhenAll to fetch them all in parallel instead of one by one. In case window is closed while fetching is in progress CancellationToken is used to cancel it all.

### DIFF
--- a/src/NuGetForUnity/Editor/Configuration/ConfigurationManager.cs
+++ b/src/NuGetForUnity/Editor/Configuration/ConfigurationManager.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.IO;
+using System.Net;
 using System.Runtime.CompilerServices;
 using System.Threading;
 using System.Threading.Tasks;
@@ -235,6 +236,7 @@ namespace NugetForUnity.Configuration
         /// <param name="includePrerelease">True to include pre-release packages (alpha, beta, etc).</param>
         /// <param name="targetFrameworks">The specific frameworks to target?.</param>
         /// <param name="versionConstraints">The version constraints?.</param>
+        /// <param name="token">The cancellation token?.</param>
         /// <returns>A list of all updates available.</returns>
         [NotNull]
         [ItemNotNull]
@@ -242,9 +244,20 @@ namespace NugetForUnity.Configuration
             [NotNull] IEnumerable<INugetPackage> packagesToUpdate,
             bool includePrerelease = false,
             string targetFrameworks = "",
-            string versionConstraints = "")
+            string versionConstraints = "",
+            CancellationToken token = default)
         {
-            return ActivePackageSource.GetUpdates(packagesToUpdate, includePrerelease, targetFrameworks, versionConstraints);
+            // Fetching package updates often requires a lot of requests so we increase the limit to speed them up.
+            var oldConnectionLimit = ServicePointManager.DefaultConnectionLimit;
+            try
+            {
+                ServicePointManager.DefaultConnectionLimit = 20;
+                return ActivePackageSource.GetUpdates(packagesToUpdate, includePrerelease, targetFrameworks, versionConstraints, token);
+            }
+            finally
+            {
+                ServicePointManager.DefaultConnectionLimit = oldConnectionLimit;
+            }
         }
 
         /// <inheritdoc cref="INugetPackageSource.GetSpecificPackage(INugetPackageIdentifier)" />
--- a/src/NuGetForUnity/Editor/Configuration/ConfigurationManager.cs
+++ b/src/NuGetForUnity/Editor/Configuration/ConfigurationManager.cs
@@ -247,17 +247,7 @@ namespace NugetForUnity.Configuration
             string versionConstraints = "",
             CancellationToken token = default)
         {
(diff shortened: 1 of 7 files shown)


popara96 (approved):

src/NuGetForUnity/Editor/PackageSource/NugetPackageSourceV3.cs
```diff
@@ -242,17 +243,22 @@ public List<INugetPackage> GetUpdates(
...
                                             searchQueryBuilder.ToString(),
                                             0,
                                             0,
                                             includePrerelease,
-                                            CancellationToken.None)
-                                        .ConfigureAwait(false));
+                                            token)
+                                        .ConfigureAwait(false);
+
+                                fetchTasks.Add(SearchPackageAsync());
```
JoC0de, line 256: No need to create a anonym function
```suggestion
                                fetchTasks.Add(ApiClient.SearchPackageAsync(
                                            this,
                                            searchQueryBuilder.ToString(),
                                            0,
                                            0,
                                            includePrerelease,
                                            token));
```

src/NuGetForUnity/Editor/PackageSource/NugetPackageSourceV3.cs
```diff
@@ -242,17 +243,22 @@ public List<INugetPackage> GetUpdates(
...
                             }
 
+                            var nestedUpdates = await Task.WhenAll(fetchTasks).ConfigureAwait(false);
+                            var updates = nestedUpdates.SelectMany(r => r).ToList();
```
JoC0de, line 260:
```suggestion
                            var updates = nestedUpdates.SelectMany(packages => packages).ToList();
```

src/NuGetForUnity/Editor/Ui/NugetWindow.cs
```diff
@@ -554,7 +611,8 @@ private void DrawUpdates()
             }
             else
             {
-                DrawNoDataAvailableInfo("There are no updates available!");
+                var msg = fetchingUpdates ? "Fetching package updates. Please wait..." : "There are no updates available!";
+                DrawNoDataAvailableInfo(msg);
```
JoC0de, line 615:
```suggestion
                var message = fetchingUpdates ? "Fetching package updates. Please wait..." : "There are no updates available!";
                DrawNoDataAvailableInfo(message);
```

src/NuGetForUnity/Editor/Configuration/ConfigurationManager.cs
```diff
@@ -235,16 +236,28 @@ public static Task<List<INugetPackage>> SearchAsync(
...
-            string versionConstraints = "")
+            string versionConstraints = "",
+            CancellationToken token = default)
         {
-            return ActivePackageSource.GetUpdates(packagesToUpdate, includePrerelease, targetFrameworks, versionConstraints);
+            // Fetching package updates often requires a lot of requests so we increase the limit to speed them up.
+            var oldConnectionLimit = ServicePointManager.DefaultConnectionLimit;
+            try
+            {
+                ServicePointManager.DefaultConnectionLimit = 20;
```
JoC0de, line 254: This will only affect fetching from nuget source with V2 for V3 we would need to change the setting on the `HttpClient`. Not sure if we should also change it for V3.

JoC0de (changes requested):

src/NuGetForUnity/Editor/Configuration/ConfigurationManager.cs
igor84: I tried setting MaxConnectionsPerServer on HttpHandler that we use but I got NotImplementedException. It seems HttpClient in Unity Editor is using MonoHttpHandler which doesn't support that. In my testing this is the only option I found that actually sped up fetching updates over V3.

src/NuGetForUnity/Editor/PackageSource/NugetPackageSourceV3.cs
igor84: Don't we need ConfigureAwait on each task? I am never sure about that... I'll test it like this on Monday and accept the change if it works out.

JoC0de: When you don't have a `await` you don't wait for the task to be completed so you don't need to define on witch thread the code after the await (run when the task finished) should run.

src/NuGetForUnity/Editor/Configuration/ConfigurationManager.cs
JoC0de: ok could be that Mono even uses `ServicePointManager.DefaultConnectionLimit` internally

igor84:
I had to add try catch in GetUpdates Task.Run because Unity would sometimes get stuck if OperationInterruptedException was thrown within.
I also had to remove setting ServicePointManager.DefaultConnectionLimit because it does nothing unless set very early after Domain reload. For example it works when I set it from a method marked with [UnityEditor.InitializeOnLoadMethod] attribute. I found that documentation for it says "Set your preferred value once, when the AppDomain loads".

Issue #672, opened 2024-10-25
Unity 6 compatibility?


MarkHelsinki:
Am I right in thinking that this isn't yet compatible with Unity 6?

Seems to be an issue with ZipFile not being recognised.

popara96:
Can you add more details to this? I can't seem to find any issues while using it in Unity 6.

r8420:
> Am I right in thinking that this isn't yet compatible with Unity 6?
> 
> Seems to be an issue with ZipFile not being recognised.

I had the same problem which prevented me from attaching the debugger. For me the solution was to manually delete the nuget plugin folder and reinstalling the latest version. After that I reopened Visual Studio and everything worked fine again.

MarkHelsinki:
![Screenshot_261](https://github.com/user-attachments/assets/a7e69136-2934-47d3-8537-ff4b0ab87282)
Just did a clean install. The Library cache is also recent, meaning it should be clean. This was installed via Package Manager and the URL.

MarkHelsinki:
Okay. Deleted the folder: com.unity.nuget.mono-cecil in the package cache and it reinstalled without errors. Case closed :)

MarkHelsinki:
Gosh, I spoke too soon. It seemed to be fine, meaning it had compiled and didn't show the errors, but the exact same compile errors are back again.

gigacee:
same error here, Unity 6000.0.29f1

gigacee:
Set "Editor Assemblies Compatibility Level" to ".NET Standard" and errors are disappear.

<img width="424" alt="image" src="https://github.com/user-attachments/assets/10c28a57-d628-42a1-9b68-69ea7b979d91">

JoC0de:
I also got a situation where I had the errors. But thy are only in visual studio / raider and so they don't affect Unity. You can even deactivate that Unity creates Project files for UPM packages -> so you don't get the errors in the IDE.

yoneyamaharuto:
I have same problem in unity 6000.0.28 rider

JoC0de:
Just to clarify this are only compiler errors shown in the IDE -> anything will still work in Unity. I would recommend to just deactivate the "Project files generation for UPM packages" so you only see your own code in the IDE (so there will be no errors).

TheXRMonk:
[...]

PR #724, opened 2025-07-14
Disable autoReferenced in NuGetForUnity.asmdef

apkd:
This is a proposed workaround for IDE-side build problems caused by #672.

The `autoReferenced` setting causes the NuGetForUnity assembly to be added as a dependency to the Assembly-CSharp assembly, making it difficult to use certain IDE features (e.g., Rider's IL viewer, which requires the project to be built).

This change implies breaking backwards compatibility for users that rely on accessing the NuGetForUnity assembly from Assembly-CSharp and friends. I think it's a bit of an anti-pattern for an editor tool anyway, and extending NuGetForUnity can always be achieved via direct assembly references.

### DIFF
--- a/src/NuGetForUnity/Editor/NuGetForUnity.asmdef
+++ b/src/NuGetForUnity/Editor/NuGetForUnity.asmdef
@@ -11,7 +11,7 @@
     "precompiledReferences": [
         "NuGetForUnity.PluginAPI.dll"
     ],
-    "autoReferenced": true,
+    "autoReferenced": false,
     "defineConstraints": [],
     "versionDefines": [],
     "noEngineReferences": false


TheXRMonk:
Wow, just in time! I was about to add my two cents on why "just deactivate the 'Project files generation for UPM packages'" is not viable for anyone who use writable packages as submodules or otherwise. <3

apkd:
After some testing, it looks like the correct fix for #672 is to remove or empty the `csc.rsp` file.

The compilation errors are actually proceded by the warning:
```
4>Microsoft.Common.CurrentVersion.targets(2433,5): Warning MSB3243 : No way to resolve conflict between "System.IO.Compression, Version=4.2.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089" and "System.IO.Compression, Version=4.1.3.0, Culture=neutral, PublicKeyToken=b77a5c561934e089". Choosing "System.IO.Compression, Version=4.2.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089" arbitrarily.
```

Looks like `System.IO.Compression` is already brought in by default, and referencing it using `csc.rsp` causes a conflict. Not sure if removing `csc.rsp` breaks compatibility with earlier Unity/IDE versions. Can someone check? I'm assuming it was needed for something at some point? 

I still think disabling `autoReferenced` seems like a good idea, so I'll leave this pull request around for now.

JoC0de (approved):
I think it is even expected for editor plugins to not be auto-referenced.
The `csc.rsp` file is at least needed for the plugin to work in Unity 2018 if we increase the minimum required Unity version it could be possible that we don't need it.

PR #729, opened 2025-09-05
Add win-arm64, linux-arm64, android-x64, android-arm, ios-x64 to NativeRuntimeSettings.DefaultSettings

neuecc:
First of all, thank you for the wonderful library. I am actively recommending its use in my OSS projects as well.

While testing libraries that include native runtimes, I confirmed that some packages are not being imported.

image

This is caused by the absence of entries in NativeRuntimeSettings, and I have also confirmed that there are no issues if overwritten with the corresponding json.
https://github.com/Cysharp/NativeCompressions/blob/6123b5a/sandbox/UnityApp/ProjectSettings/Packages/com.github-glitchenzo.nugetforunity/NativeRuntimeSettings.json

This is a proposal to include support for these (win-arm64, etc.) in the default settings.


JoC0de (approved):

JoC0de:
@copilot apply changes based on the comments in [this thread](https://github.com/GlitchEnzo/NuGetForUnity/pull/729#pullrequestreview-3674607088)

PR #743, opened 2026-01-08
Fixed nuget packages get lost with Unity 6

tgoessler:
In Unity 6 it can happen that multiple asset import worker processes get started at the same time then this will lead to undefined behavior because session data is new for each worker process, so it will execute the code in parallel multiple times.


JoC0de (approved):
Even if the `if` will not work when NuGetForUnity is installed using `.unitypackage` install method it is ok.

PR #746, opened 2026-02-22
Menu Items in Proper place (imo)

martincalander:
I feel like this is better place to store the menuitems..


popara96:
This seems a bit convoluted. 'Window' menu item is already packed with loads of Unity options, and a lot of 3rd party plugins also stuff their menus there. What you could do though is make the root of this menu item be configurable (i.e. make it 'NuGet' by default but able to be changed within NugetPreferences).

src/NuGetForUnity/Editor/Configuration/NugetConfigFile.cs
```diff
@@ -202,6 +212,16 @@ public string RelativePackagesConfigDirectoryPath
         /// </summary>
         internal bool KeepingPdbFiles { get; set; }
 
+        /// <summary>
+        ///     Gets or sets the root menu path used for NuGet menu items.
+        /// </summary>
+        [NotNull]
+        public string MenuRoot
```
JoC0de:
```suggestion
        internal string MenuRoot
```

martincalander:
Bump. I've now made the Menu Item placement customizable, and defaults to the original root/NuGet.

PR #752, opened 2026-05-02
Add support for ignored package dependencies

Tr0sT:
## Summary
  - add `ignoredPackageDependencies` configuration and Project Settings UI for dependencies that should be skipped when installed transitively
  - skip ignored dependencies during install/restore and remove stale implicit entries from `packages.config`
  - add an Installed tab action to uninstall an implicit package and add it to the ignore list

This is useful for Unity projects where some NuGet dependencies are already provided by another plugin or SDK installed outside of NuGetForUnity. Without an ignore option, NuGetForUnity installs the transitive dependency again, which can create duplicate assemblies in the project and lead to compilation errors, type conflicts, or ambiguous runtime behavior. Ignoring selected implicit dependencies lets the project keep using the externally provided copy while still allowing NuGetForUnity to restore and manage the packages that actually need to be installed.


Tr0sT:
@GlitchEnzo

PR #754, opened 2026-05-22
Fix analyzer DLL import settings before Unity refresh

rafay-pk:
## Summary

Fixes NuGet packages with Roslyn analyzer/source-generator DLLs, such as `MemoryPack.Generator`, being imported by Unity as normal plugin assemblies before NuGetForUnity can label them as analyzers.

Unity requires analyzer DLLs to have plugin platforms disabled and the `RoslynAnalyzer` asset label set. Without the `.dll.meta` file in place before `AssetDatabase.Refresh`, Unity can try to load analyzer assemblies as runtime assemblies and fail on references such as `Microsoft.CodeAnalysis`.

## Changes

- Add shared analyzer `.dll.meta` generation in `AnalyzerHelper`
- Write initial analyzer import settings during package install before Unity refreshes assets
- Reuse the same analyzer meta writer from the CLI restore path
- Guard Roslyn-version selection when no analyzer version is supported by the current Unity version
- Update analyzer install test to assert metadata exists before `AssetDatabase.Refresh`

## Verification

- Confirmed locally in Unity that installing `MemoryPack`/`MemoryPack.Generator` works without the `Microsoft.CodeAnalysis` plugin load errors
- `dotnet build src\NuGetForUnity.Cli\NuGetForUnity.Cli.sln`
- `dotnet test src\NuGetForUnity.Cli\NuGetForUnity.Cli.sln --no-build`
- `dotnet build src\NuGetForUnity.PluginAPI\NuGetForUnity.PluginAPI.sln`


igor84:
Should this change include removal of RoslynAnalyzer code from NugetAssetPostprocessor?

rafay-pk:
It is an early/import-safe bootstrap, not a replacement for the postprocessor. The new installer-side `.meta` generation fixes the timing problem: Unity sees correct analyzer import settings before the first refresh/import. But the postprocessor is still useful as the normal Unity-side enforcement path:
- It handles analyzer DLLs that appear without prewritten .meta files, including manual changes or older installs.
- It keeps existing behavior for reimports and asset database processing.
- It adds the `NuGetForUnity` processed label and applies importer settings through Unity APIs, which is still the most authoritative path once Unity has an importer.
- It avoids making CLI/editor extraction metadata the only source of truth.

